Issue #4, opened 2017-05-05
add support for pinboard


Finkregh:
it would be super awesome if there would be support for [pinboard](https://pinboard.in/tour/); they also have the paid feature to archive your bookmarks

pirate:
@Finkregh I don't have a pinboard account, can you make a [gist](https://gist.github.com) containing a typical pinboard export?  I just need to know what format it uses in order to tweak my parsing regex.

pigmonkey:
I dump my Pinboard bookmarks as JSON. It looks like this:

```
[
    {"href":"https:\/\/github.com\/pirate\/pocket-archive-stream","description":"GitHub - pirate\/pocket-archive-stream: \ud83d\uddc4 Save an archived copy of all websites starred using Pocket. Outputs browseable html.","extended":"","meta":"6c8e6537a8e608125ae73eb1ae1edf69","hash":"ed18b7d35889b1395cf13d0bc69f1b80","time":"2017-05-05T20:31:24Z","shared":"no","toread":"no","tags":"archive python"},
    ...
]
```

The API also supports XML. It looks like this:

```
<?xml version="1.0" encoding="UTF-8" ?>
<posts user="pigmonkey">
<post href="https://github.com/pirate/pocket-archive-stream" time="2017-05-05T20:31:24Z" description="GitHub - pirate/pocket-archive-stream: 🗄 Save an archived copy of all websites starred using Pocket. Outputs browseable html." extended="" tag="archive python" hash="ed18b7d35889b1395cf13d0bc69f1b80" meta="6c8e6537a8e608125ae73eb1ae1edf69"  shared="no"  />
...
</posts>
```

PR #7, opened 2017-05-05
Add pinboard support

sconeyard:
I renamed some variable names to be more neutral towards new services. 
In general I added a new method which parses the json dump you can retrieve from pinboard. Nothing fancy, just some quick changes. 

Regards #4


pirate:
+1 thanks!

PR #9, opened 2017-05-06
Fix default parameters parsing and typo in logs when no links found

Neavend:
Changed the old way parameters were parsed as this was not taking the default "ril_export.html" file when none were precised, nor the "pocket" option


pirate:
Thanks!

PR #10, opened 2017-05-06
Update README.md

geirolav:
Updated syntax ( ' instead of " for section, and enclosing parameter switch in "" ) for creating alias on OSX. Took a while for a beginner like me to discover/get right, might be a problem for others as well if they are not well versed in these things. Tested on MacOS Sierra. Great project :-)


pirate:
Thanks for the suggestion, but why change this?  I tested the existing version and it was working for me on Sierra, were you having issues with it?

pirate:
Oh nevermind, I see what happened, it was only working for single args due to globbing. Thanks!

PR #13, opened 2017-05-08
Update README.md

Miladiir:
The extra backslash is unnecessary and breaks the script. PDFs and screenshots are not generated.


Miladiir:
Some unix distros like Debian Jessie do not provide a recent enough version of Python3. I think 3.5 is required atleast, but Debian currently only provides up to 3.4.x.

The readme was updated to reflect the required package versions.

pirate:
Thanks, we could also toss this in there somewhere:

```bash
add-apt-repository ppa:fkrull/deadsnakes
apt update && apt install python3.6
```

Issue #18, opened 2017-05-29
Support parsing links from browser bookmark exports


pirate:
Google Chrome supports exporting links from the bookmark manager as an HTML file.

It'll be trivial to add a parser for bookmark exports, it just involves adding a new service type and parser.

wlhlm:
The HTML export is supported by many browsers and bookmarking services.

- [Netscape Bookmark File Format](https://msdn.microsoft.com/en-us/library/aa753582(v=vs.85).aspx)
- [Sample](https://gist.github.com/jgarber623/cdc8e2fa1cbcb6889872)
- [HTML->JSON Converter written in Python](https://github.com/jhh/netscape-bookmark-converter).

Firefox at least also supports exporting bookmarks as JSON ("backup"). I haven't found documentation about the format, but should be fairly straight forward to parse.

wlhlm:
@ilvar Thank you!

PR #20, opened 2017-06-08
#18 - Parse browser bookmarks

ilvar:
Added [BeautifulSoup4](https://www.crummy.com/software/BeautifulSoup/bs4/doc/) to parse browser bookmarks dumps (issue #18).


pirate:
Thanks for the contribution!   I think I'd rather keep it just with regex so we can stay at 0 dependencies though....

pirate:
I'll merge it for now and work on refactoring to regex later.

PR #21, opened 2017-06-11
Removed BeautifulSoup from dependencies

ilvar:


pirate:
What a boss.  Thanks so much for doing this!

PR #33, opened 2017-07-05
Use template strings for substitution in HTML output

bardisty:
`str.format()` can only use substitutions identified by braces (`{` and `}`). This has the potential to conflict with other code in the HTML template, such as CSS or JavaScript.

Template strings can use substitutions identified by `$` or `${}`, e.g.: `$identifier` or `${identifier}`. These substitutions won't conflict with CSS or JavaScript, allowing users to write HTML templates that don't require double braces anywhere there's a substitution conflict. This is especially useful when one is using a build tool to generate the final CSS/JavaScript/HTML.
    
https://docs.python.org/3/library/string.html#template-strings

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,6 @@ env CHROME_BINARY=google-chrome-stable RESOLUTION=1440,900 FETCH_PDF=False ./arc
 
 To tweak the outputted html index file's look and feel, just copy the files in `templates/` somewhere else and edit away.  Use the two index config variables above to point the script to your new custom template files. 
 
-The templates use format strings (not a proper templating engine like jinja2), which is why the CSS is double-bracketed `{{...}}`.
-
 ## Publishing Your Archive
 
 The archive produced by `./archive.py` is suitable for serving on any provider that 
--- a/index.py
+++ b/index.py
@@ -1,5 +1,6 @@
 import os
 from datetime import datetime
+from string import Template
 
 from config import INDEX_TEMPLATE, INDEX_ROW_TEMPLATE
 from parse import derived_link_info
@@ -16,7 +17,7 @@ def dump_index(links, service):
         link_html = f.read()
 
     article_rows = '\n'.join(
-        link_html.format(**derived_link_info(link)) for link in links
+        Template(link_html).substitute(**derived_link_info(link)) for link in links
     )
 
     template_vars = {
@@ -27,4 +28,4 @@ def dump_index(links, service):
     }
 
     with open(os.path.join(service, 'index.html'), 'w', encoding='utf-8') as f:
-        f.write(index_html.format(**template_vars))
+        f.write(Template(index_html).substitute(template_vars))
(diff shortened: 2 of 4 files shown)


bardisty:
After some further testing I've found some JavaScript can still cause conflicts (libraries such as jquery):

```text
[+] [2017-07-05 04:29:06] Starting archive from /home/bah/var/pinboard/pinboard.json export file.
  File "./archive.py", line 88, in <module>
    create_archive(export_file, service=export_type, resume=resume_from)
  File "./archive.py", line 55, in create_archive
    dump_index(links, service)
  File "/home/bah/src/gh/bookmark-archiver/index.py", line 31, in dump_index
    f.write(Template(index_html).substitute(template_vars))
  File "/usr/lib/python3.6/string.py", line 126, in substitute
    return self.pattern.sub(convert, self.template)
  File "/usr/lib/python3.6/string.py", line 123, in convert
    self._invalid(mo)
  File "/usr/lib/python3.6/string.py", line 99, in _invalid
    (lineno, colno))
ValueError: Invalid placeholder in string: line 1, col 9377
```

This can be bypassed by using `safe_substitute()` over `substitute()`, but I haven't comitted this as it has a small caveat: 

```diff
- Template(link_html).substitute(**derived_link_info(link)) for link in links
+ Template(link_html).safe_substitute(**derived_link_info(link)) for link in links

- f.write(Template(index_html).substitute(template_vars))
+ f.write(Template(index_html).safe_substitute(template_vars))
```

> Like substitute(), except that if placeholders are missing from mapping and kwds, instead of raising a KeyError exception, the original placeholder will appear in the resulting string intact. Also, unlike with substitute(), any other appearances of the $ will simply return $ instead of raising ValueError.

> While other exceptions may still occur, this method is called “safe” because substitutions always tries to return a usable string instead of raising an exception. In another sense, safe_substitute() may be anything other than safe, since it will silently ignore malformed templates containing dangling delimiters, unmatched braces, or placeholders that are not valid Python identifiers.

via https://docs.python.org/3/library/string.html#string.Template.safe_substitute

If `safe_substitute()` is used, any conflicting JavaScript will simply be inserted as strings in the template, which is what we'd want and would allow users to use any JavaScript they want (in my case, date-fns for relative dates e.g., "1 day ago"). However that means ignoring potentially malformed templates. I figure that will most likely involve not seeing any errors if a placeholder contains a typo, is obsolete, has been renamed, etc.

As this is something that would likely only affect users using custom templates, perhaps making it an option would be a better route? Default would be to use `substitute()`, and users with custom templates could opt in to use `safe_substitute()` instead. Not sure what the option would be called, maybe `IGNORE_TEMPLATE_ERRORS`, `IGNORE_INVALID_PLACEHOLDERS`, or whatever makes the most sense. Seems like a decent route short of using a template engine where conflicting placeholders isn't an issue.

pirate:
Nice, I forgot Template strings existed in Python, haven't seen them used in years.  Thanks for the contribution!  This is a much-welcome improvement over double-bracing everything in CSS.

I'm considering moving to a Flask backend with jinja2 when I add `server.py` which provides a search backend and a GUI to submit new links.  (`archive.py` will still exist independently as a static-html generator)

I'm ok just having `substitute()` for the static generator script for now, and more advanced users can customize the jinja templates used by the server.

PR #45, opened 2017-09-04
ensure encoding check is case-insensitive

klauern:
encoding will fail if it isn't 'UTF-8', resulting in encodings of 'utf-8' to fail, too... This normalizes it.


pirate:
Goid point!  But I think you meant `.upper()` or `.lower() != 'utf-8'` here.

klauern:
so sorry about that.  fixed...

pirate:
No worries, thanks for the fix!

PR #47, opened 2017-11-02
Fix usage of ARCHIVE_DIR if an exisiting archive is updated

m-rossi:


pirate:
Good catch, thank you!

Issue #39, opened 2017-08-03
Link Parsing: Add support for plain-text list of URLs


nodiscc:
Currently the input file is required to be in the Netscape HTML export format. I have lists of URLs I'd like to archive, in plain-text format. Converting them to HTML should be doable, but I wish bookmark-archiver would accept plain text files, such as

`./archive.py urls.txt`

It would be reasonable to filter the file, retaining actual `http(s)?://...` URLs

pirate:
Yeah I thought about this, but BA doesn't just take URLS it needs titles and timestamps too to work properly.

nodiscc:
I could write a basic txt -> Netscape HTML converter like

```bash
date=$(date +%s)
textfile="links.txt"
urllist=$(egrep --only-matching 'https?://\S+\s' "$textfile") #find all URLs in file
for pageurl in $urllist; do
    date=(( $date + 1 ))
    pagetitle=$(wget "$pageurl" -q -O - | awk -vRS="</title>" '/<title>/{gsub(/.*<title>|\n+/,"");print;exit}') #extract webpage title
    if [ "$pagetitle" = "" ]; then pagetitle="$pageurl"; fi
    htmlentry="<DT><A HREF=\"$pageurl\" ADD_DATE=\"$date\">$pagetitle</A>"
    echo "$htmlentry" >> output-netscape.html
done
```

(just a quick bash proof of concept but I could try to rewrite it in Python)

and let BA work from there . This would add a first pass to find all links in the text file, gather their page titles, and write a BA compatible HTML file. The downside is that it would double the number of page downloads, maybe they can be cached? At least it would solve my problem without requiring refactoring.

Another option is to generate an HTML file without page titles :/ 

    <DT><A HREF=\"$pageurl\" ADD_DATE=\"$date\">$pageurl</A>


What do you think?

pirate:
Actually adding timestamps isn't needed, the script handles deduping them already.  If you just set all the timestamps to 1, it will parse them as 1.1, 1.2, 1.3, 1.4, ...

The tricky issue is titles, but I predict your parser will work for most sites.

sbrl:
Great script, @nodiscc! I've taken the liberty of improving it slightly. Here's a gist of the result of 1/2 hour of tinkering: https://gist.github.com/sbrl/95d1c23e18def900aeca35c2f2e57f24

pirate:
[...]

PR #59, opened 2017-12-28
Add a parser for a list of links (see issue #39)

frinkelpi:
See https://github.com/pirate/bookmark-archiver/pull/58 and the comments therein. Thanks, and don't hesitate if you have any question on the changes.


issmirnov:
This is an awesome feature. Does this project still hold your interest and would you find time to fix the merge conflicts?

issmirnov:
Ah, interesting. Yes, this makes sense. I just created a small `bookmarks.html` with fake data, ran into this issue.

What ID scheme did you decide to use? Domains + page path hash, or something else?

pirate:
Full url actually, including scheme, it's what the Internet Archive uses and I'd rather archive two similar urls separately than merge two different pages into one by accident.

I'll probably drop trailing slashes, but http and https would be different links.  Most sites redirect to https now anyways, so I doubt it's going to be an issue.

pirate:
I recommend using @nodiscc's [script to convert a plain-text list](https://github.com/pirate/bookmark-archiver/issues/39#issuecomment-320066412) into a valid netscape html format for now.

issmirnov:
Awesome, that also makes my personal use case way simpler. Planning to archive all links on my wiki, so having a deterministic prefix is ideal.

If there are any subtasks you can farm out, I'd be happy to take a stab at them.

pirate:
That would be awesome, thanks @issmirnov!  This is a good first ticket to get familiar with the codebase: https://github.com/pirate/bookmark-archiver/issues/38

It involves adding a couple new options in `config.py`, and adding some logic to skip certain urls when creating the index and archive.

pirate:
I'm going to close this because I'm removing timestamps as the unique key and using url hashes instead in the new layout as mentioned here: 
https://github.com/pirate/bookmark-archiver/issues/120

Once unique ordering is established separately from timestamps, link parsing from a plain text list of links will be trivial and I'll make sure I push it with the new layout changes.

sbrl:
Awesome, thanks! I guess it'll be in the next release?

nodiscc:
Using URL hashes is very nice and i drew the same conclusion when working on my own archiving tool (now unmaintained). Good luck with the next release

PR #65, opened 2018-01-20
WIP: Create python package from repository

edoput:
This will create a python package installable using *pip*.

The package can be later published on pypi for easier access.

Before merging I would squash everything into one commit if approved.

### Scripts

the installation provide an `archive` command that will be available from the shell and will execute the `archive.py` script

### Setup

The important part is the `setup.py` file as it contains metadata and instructions for pip.

I filled it with the information I could find and it should be ok but as you are the author please review it.

### config.py

As this file is considered editable by the user maybe we should move it somewhere suitable (`~/.config/bookmark-archiver/config.py`) and access it at runtime.


pirate:
Thanks for this work!

I've been thinking about a good way to do config for a while now, and I think the logical next step is something like YAML, JSON, or another easy-to-parse config format.

edoput:
This is a first step to move the configuraiton out of the package. I've choose to go with a ini file as it can be parsed with the python standard library and the parser class supports the precedence user > system natively.

I added two configuration file, one system-wide and one in `~/.config/bookmarks-archiver/user.conf`.

There is no default for the [`DIR` configuration value](https://github.com/edoput/bookmark-archiver/blob/master/conf/archiver.conf#L6). As it is used in other values [(`HTML_FOLDER`, `ARCHIVE_FOLDER`)](https://github.com/pirate/bookmark-archiver/blob/master/config.py#L26) I would like to know what it should be when configured.

edoput:
I feel this is coming together nicely and it will provide some improvements.

- the project can be installed from pip
- the `archive` script is linked in `/usr/bin/archive`
- the templates are distributed with the project but we can move them out and use them as _plugins_ just like the `sphinx` project uses plugins for it's themes (e.g. [guzzle theme](https://github.com/guzzle/guzzle_sphinx_theme)
- there is a configuration file for system-wide and one for user configuration. The `ini` format is simple for user to parse and it ships with python
- I've added a `archive-config` script to dump the configuration to screen so it can be inspected

pirate:
Cool, can we group the INI config options by archive method, e.g.:

[wget]

[video]

[audio]

[screenshot]

[pdf]

etc.

edoput:
Yes it shouldn't be a problem.

Right now the __subsections__, e.g. `[website:pdf]` are only an aestetic matter. I have no problem with renaming it to `[pdf]` if you think it's important. I guess a plugin that provides an archive method can embed it's configuration in a section.

As an example a __screenshot__ plugin can use the `[screenshot]` section.

Let me know if you feel we are missing something.

pirate:
I haven't forgotten about this, but I probably won't have time to fully review it and merge it anytime soon.

It includes a lot of changes so I want to make sure I thoroughly test it on mac and linux before merging.  Currently spending all my time on https://oddslingers.com, but I'll post back here once I get a chance to test this branch.

edoput:
Let's move when we are ready, I'm in no hurry and would like to not break anything

pirate:
Fyi just to keep you in the loop, this is my next planned refactor for BA: https://docs.google.com/presentation/d/17gPkY6chAzDDXCce1QkTdRwl2nChESdjfibal-n8TgM/edit?usp=sharing

I'm adding a django backend, sql database, and scheduler in order to build a GUI that lets people run scheduled imports from multiple sources, but I still want to keep the cli version available for free forever.

I think the pypi package is the perfect solution for initial setup option #3 in the slides.

edoput:
> looking at your pr, archive.py and and archive.config can actually be included on module. the command line can be configured using setup.py's entry point.

They could be included but then you are losing a separation of concern. Scripts are executable the module is a library.

> e: also with multiple entry point it can also run server, i.e bookmark-archive for command line and bookmark-archive-server for server

yes you are right but I wouldn't use a command to start the server, it should be a process started by your init daemon somehow. Flask is just a framework but Python supports something called wsgi as an interface with webservers, what you would like is to have a wsgi deamon load you app.

@pirate I thought a lot about configuration files and I got to the point where I am 100% sure that pip shouldn't write to system configuration paths like `/etc/`. This is handled by packages and distributions and I will remove the creation of system wide configuration during installation.

Cheers

rachmadaniHaryono:
>>looking at your pr, archive.py and and archive.config can actually be included on module. the command line can be configured using setup.py's entry point.

>They could be included but then you are losing a separation of concern. Scripts are executable the module is a library.

https://stackoverflow.com/questions/2996110/what-is-the-difference-between-a-module-and-a-script-in-python

based on this https://packaging.python.org/tutorials/distributing-packages/#scripts

>Although setup() supports a scripts keyword for pointing to pre-made scripts to install, the recommended approach to achieve cross-platform compatibility is to use console_scripts entry points (see below).

but if cross-platorm compatibility is no concern, i think it can be skipped

---

>>e: also with multiple entry point it can also run server, i.e bookmark-archive for command line and bookmark-archive-server for server

>yes you are right but I wouldn't use a command to start the server, it should be a process started by your init daemon somehow. Flask is just a framework but Python supports something called wsgi as an interface with webservers, what you would like is to have a wsgi deamon load you app.

i actually got this idea after looking at search.py on the pr and this feature at buku

https://github.com/jarun/Buku/pull/257

it need more work to deploy (one from this http://flask.pocoo.org/docs/0.12/deploying/) but user can run the local host easily

pirate:
I'm removing flask FYI in favor of django, see the `server` branch.

The django will be started via `runserver` or `wsgi`.

Django will be used to manage the migrations for the SQLite database as well.

edoput:
In the [config section](https://github.com/pirate/bookmark-archiver/pull/65/files#diff-375df057e13447167d03f9c364db1b39R52) for the templates I added this but now I don't know if it's cool with the new django approach, maybe I should scrap it

pirate:
I can make ConfigParser work with Django, but I cant promise I'll be able to get the django alpha branch in a usable state within the next 6 weeks.

Whether you want to wait till I'm done with django in a few months, or work on it as-is and let me handle the merge later is up to you.

edoput:
Ok so, this is working on my side but here's the catch, using the templates is going to be a little strange.

I would prefer to remove this template configuration at all as I don't really know if it's compatible with django but I guess it could be.

- the template configuration is not using the configuration file
- templates are stored in the package as `package_data`
- static are stored in the package as `package_data`
- templates are retrieved using `pkg_resources`, this means that we can define an _interface_ for template packages. I don't know if this is compatible with django tempates.
- static are retrieved using `pkg_resource`, again an _interface_ is possible.

### templates packages

This means that when using `pkg_resource.resource_filename` we can pass a `name` for a package containing the following data.

This name could be the value in the configuration file, e.g:
```ini
[templates]
package=foo
```
 
And then in the package foo the directory templates contains the following tree

```
foo/templates/
├── index.html
├── index_row.html
├── link_index_fancy.html
├── link_index.html
└── static
    ├── archive.png
    ├── external.png
    └── spinner.gif

1 directory, 7 files
```

pirate:
Django can easily import any user-supplied template dir from the config file, without it needing to be a package.  We can tell django to use its own built-in templates folder first, but prefer any templates found in a user-provided folder if they exist.  Inside the template files I'll also need to translate the syntax from python string templates to django handlebars, but that's not too difficult.

`server/settings.py`:
```python
from bookmark_archiver.config import TEMPLATES_DIR

TEMPLATES = [
    {
        'BACKEND': 'django.template.backends.django.DjangoTemplates',
        'DIRS': ['templates', TEMPLATES_DIR],
        'APP_DIRS': True,
        'OPTIONS': {
            'context_processors': [
                'django.template.context_processors.debug',
                'django.template.context_processors.request',
                'django.contrib.auth.context_processors.auth',
                'django.contrib.messages.context_processors.messages',
            ],
        },
    },
]
```

rachmadaniHaryono:
does this line correct?

https://github.com/edoput/bookmark-archiver/blob/16c0e4196c544e738ac32ca7485dd6c00c60cdc8/bin/archive#L94

or this from actual upstream repo

https://github.com/pirate/bookmark-archiver/blob/a532d1154910577f7954897c874e052dda70e73f/archive#L119

for example after installing it on virtualenv, run `archive --help` and 

(Pdb++) sys.argv
['/home/<user>/envs/bookmark-archiver/bin/archive', '--help']
(Pdb++)
['/home/q/envs/bookmark-archiver/bin/archive', '--help']
(Pdb++) set(sys.argv).intersection('-h', '--help', 'help')
set()

which would be evaluated as false.

add another bracket?

(Pdb++) set(sys.argv).intersection(('-h', '--help', 'help'))
{'--help'}

pirate:
Good catch, yeah it just needs another bracket.

rachmadaniHaryono:
if it is turned into module, it can be tested

something like

```python
main('archive', '--help')
# equivalent to 
# $ archive --help
```

on source code something similar to this

```python
def main(**args)
	args = args if args else sys.argv
	....
```

and would be better if there is proper argparse on that module.

rachmadaniHaryono:
>e: also with multiple entry point it can also run server, i.e bookmark-archive for command line and bookmark-archive-server for server

https://github.com/rachmadaniHaryono/bookmark-archiver/tree/feature/server

kinda mock up for that

- it use this pr as base
- it can be run using default flask cli, all option is also provided from flask
  - any command can be added using flask cli. see http://flask.pocoo.org/docs/1.0/cli/
  - with this, a command to import to database can be added. flask migration can be used to migrate the database. see https://flask-migrate.readthedocs.io/  
- unfortunately no index `files` link on link index page is not working. need more configuration on how to setup open directories with flask
- no db yet, using local json
- only tested with default setting from this pr, with some changes
- some difference on time handling, some text for time is more verbal
- no option to change template

pirate:
FYI all here, heres my roadmap for bookmark archiver:

Short term:
1. break up `archive/` into a plugin architecture, so that [link parsers], [index types],  and [archive methods] can be added and tested by people on their own
2. make `archive/` into a python package, so that individual BA components are usable separate from any future backend:
    `pip install bookmark-archiver; archive "..."` or   
    `import pocket from bookmark_archiver.parsers` or  
    `import archive from bookmark_archiver.archive_methods`  
3. create a django-based backend with SQLite database and migrations, full text search, link adding/removing page
...
Long term:
4. allow bookmark archiver backend instances to be linked together to share pages in a federated archive network (far in the future)
5. allow people to upload archived resources to a public federated BA network directly (e.g. upload an old ISO or video game)

I'm pretty dead-set on Django over Flask, having worked with both for many years, imo the Django ORM with it's migrations system is a much superior developer and user experience imo.

pirate:
I'm going to close this for now until the new roadmap is finalized, as it modifies the python package structure a fair amount: 
https://github.com/pirate/bookmark-archiver/issues/120

I'll definitely take inspiration from this PR in the process of implementing the new layout, but it'll likely  to take some significant work on my side before it's ready for release with the new structure.

Issue #64, opened 2018-01-12
Archive Method: Fix Medium RSS feed parsing


pirate:
```bash
➜ ~/D/C/bookmark-archiver py[archiver] ⎇ (ssl-check) ./archive.py "https://medium.com/feed/@thesquashsh"
[*] [2018-01-12 12:23:50] Downloading https://medium.com/feed/@thesquashsh > downloads/medium.com.txt
[X] No links found :(
[1]
```

aseeralmeem:
➜ ~/D/C/bookmark-archiver py[archiver] ⎇ (ssl-check) ./archive.py "https://medium.com/feed/@thesquashsh"
[*] [2018-01-12 12:23:50] Downloading https://medium.com/feed/@thesquashsh > downloads/medium.com.txt
[X] No links found :(
[1]
![14560f63-f488-4d07-a231-c09cb5df115a](https://user-images.githubusercontent.com/15217651/40426387-a1935598-5ea3-11e8-812b-6256162e4a41.jpeg)

pirate:
I've confirmed this is working on master:

```
(venv) ➜ ~/D/C/bookmark-archiver py[venv] ⎇ (cleanup) ./archive "https://medium.com/feed/@thesquashsh"                                                Sun Jun 10 22:02:37 2018
[*] [2018-06-10 22:05:04] Downloading https://medium.com/feed/@thesquashsh > output/sources/medium.com-1528682704.txt
[+] [2018-06-10 22:05:05] Adding 10 new links from output/sources/medium.com-1528682704.txt to output/index.json
[√] [2018-06-10 22:05:05] Updated main index files:
    > output/index.json
    > output/index.html
[▶] [2018-06-10 22:05:05] Updating files for 40 links in archive...
```

PR #71, opened 2018-03-14
Add support for Pinboard and Medium RSS feeds

cdzombak:
This PR adds support for Pinboard RSS feeds. It appears the current RSS parser is set up to support only Pocket feeds.

I've tested this with my own Pinboard feed: `http://feeds.pinboard.in/rss/u:cdzombak/`

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -17,6 +17,7 @@ Parsed link schema: {
 
 import re
 import json
+import xml.etree.ElementTree as etree
 
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


cdzombak:
Looking through open issues, it was simple to extend this to support Medium RSS feeds. This will fix #64 if merged. I tested this against `https://medium.com/feed/@thesquashsh` since that feed was mentioned in #64.

pirate:
Thanks so much! this is a great addition!

PR #87, opened 2018-08-07
Alphabetize config.

issmirnov:
- Alphabetize all config vars
- Factor out boolean vars to not use ".lower() == 'true'" suffixes
- Set up better tab indentation.

This is part 2 of my cleanup efforts.

### DIFF
--- a/archiver/config.py
+++ b/archiver/config.py
@@ -4,33 +4,37 @@ import shutil
 
 from subprocess import run, PIPE
 
+IS_TTY = sys.stdout.isatty()
+
+def parse_boolean_env(var, default_value):
+    """Parse string environment variables into booleans."""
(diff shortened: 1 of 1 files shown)


pirate:
This doesn't actually work though, `parse_boolean_env` doesn't return a boolean it returns a string now.  Did you mean to move the comparison inside that function?

pirate:
Thank you for your work making these changes, but after some thought I think I'm going to close it to save you time refactoring, because I'm about to change how config is handled in an upcoming release.
I'm migrating to env variables store in `.env` files and that are loaded into a django `settings.py`

I also think semantic ordering is clearer than alphabetical for config files in general because it allows for more coherent inline documentation. If people want to find a certain preference quickly they can always grep.

Issue #74, opened 2018-03-14
Uniquely identify URLs by UUID/ULID/hash of url instead of archive timestamp


cdzombak:
My Pinboard export contains several bookmarks with identical timestamps (presumably from imports from Delicious years ago).

The first time I run `archive.py`, I end up with several archive directories named like `1317249309`, `1317249309.0`, `1317249309.1`, …. These directory names correspond properly with entries in `index.json` as expected.

If I run `archive.py` a second time with the same input, it appears to rewrite `index.json`, assigning _different numerical suffixes_ to the `1317249309` timestamp. The entries in `index.json` no longer correspond with the contents of those archive directories on disk.

You can reproduce this with the following JSON file (`pinboard.json`):

```json
[{"href":"http:\/\/www.flickr.com\/groups\/photoshopsupport\/discuss\/72157600201629413\/","description":"Flickr: Discussing Index Of Topics: Compliments of LifeLive~ in Photoshop Support Group","extended":"","meta":"c9aa62c0eaa3c35a587903100870df43","hash":"8dd9951810c0eae6af67651341af5110","time":"2011-09-28T18:35:09Z","shared":"yes","toread":"no","tags":"photography photoshop retouching"},
{"href":"http:\/\/allinthehead.com\/retro\/345\/whats-in-your-utility-belt","description":"What's In Your Utility Belt? \u2014 All in the head","extended":"","meta":"746e69822f36f2e78c16fc789a7545b5","hash":"ac4d0527bca6c7d6741fee117f45f631","time":"2011-09-28T18:35:09Z","shared":"yes","toread":"no","tags":"php"},
{"href":"http:\/\/www.tyndellphotographic.com\/plasticwallet.html","description":"Plastic Wallet Boxes for Wallet sized photos","extended":"","meta":"c133eb53f29d97c35c3f31768ff7ce45","hash":"60bbf228c559518b818ed7d0ff997a69","time":"2011-09-28T18:35:09Z","shared":"yes","toread":"no","tags":"photography supply"},
{"href":"http:\/\/www.arduino.cc\/","description":"Arduino - HomePage","extended":"","meta":"a80835b5f374965f5f8a5990da6cf2be","hash":"78532ff2155cd9feeac11aba18739bdc","time":"2011-09-28T18:35:09Z","shared":"yes","toread":"no","tags":"arduino elecdiy"},
[...]

PR #96, opened 2018-09-14
fix unstable sorting between consecutive runs

fgtham:
This PR may be a possible fix for #74. It works for me (sorting remains stable), but could do with more testing.


pirate:
👍 thanks

The sorting will be improved further by https://github.com/pirate/bookmark-archiver/issues/74, but this is a great temporary fix for now.

PR #98, opened 2018-09-24
Update setup-bookmark-archiver

Strubbl:
say yes to installing dependencies via apt


Strubbl:
the second commit fixes the error:

```
  File "./archive", line 12, in <module>
    from parse import parse_links
  File "/data/archiver/parse.py", line 24, in <module>
    from util import (
  File "/data/archiver/util.py", line 6, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

pirate:
Thanks!

Issue #103, opened 2018-10-05
Archive from wallabag not working


nodh:
I've got an JSON export from the latest wallabag release (2.3.2), but bookmark-archiver won't import that:

```
Traceback (most recent call last):
  File "./archive", line 161, in <module>
    links = merge_links(archive_path=out_dir, import_path=source)
  File "./archive", line 53, in merge_links
    raw_links = parse_links(import_path)
  File "/var/www/1337.at/nextcloud-shared/bookmark-archiver/archiver/parse.py", line 52, in parse_links
    links += list(parser_func(file))
  File "/var/www/1337.at/nextcloud-shared/bookmark-archiver/archiver/parse.py", line 103, in parse_json_export
    'url': erg['href'],
KeyError: 'href'
```

pirate:
May be a while before I can take a look at this unfortunately, if you need it working ASAP you can transform the json manually to get the links into the `href` field.

Sorry for the trouble!

nodh:
Nevermind, I figured it out by myself and created a pull request that fixes this issue (at least for me).

PR #104, opened 2018-10-08
Enable importing files from wallabag

nodh:
Fix issue #103 by extracting correct fields from the wallabag export file.


pirate:
<3 thanks!

PR #107, opened 2018-10-19
Optionally import only new links

### LINKED ISSUE #95: Add a way to delete an entry from the index and archive
Occasionally I want to remove a URL from my archive. Currently this is a manual process of finding the entry in `index.json`, pulling out the timestamp, deleting the relevant lines, doing the same for `index.html`, and finally `rm -r output/archive/$timestamp`.

It would be nice if there was some slightly more automated way of doing this. Ideally I think this would be done with a final step after archiving, where the script would try to match each directory name in `output/` with a timestamp in `index.json`. If a match isn't found, the user is prompted with something like:

    1536723384 not found in bookmark index. Delete output directory? (y/n)

This may be a behavior that is only enabled by an optional config option.

f0086:
When importing a huge list of links periodically (from a big dump of
links from a bookmark service for example) with a lot of broken links,
this links will always be rechecked. To skip this, the environment
variable ONLY_NEW can be used to only import new links and skip the rest
altogether. This partially fixes #95.

### DIFF
--- a/archiver/archive.py
+++ b/archiver/archive.py
@@ -64,7 +64,7 @@ def merge_links(archive_path=OUTPUT_DIR, import_path=None, only_new=False):
         all_links = validate_links(existing_links + all_links)
     
     num_new_links = len(all_links) - len(existing_links)
-    if num_new_links:
+    if num_new_links and not only_new:
         print('[{green}+{reset}] [{}] Adding {} new links from {} to {}/index.json'.format(
             datetime.now().strftime('%Y-%m-%d %H:%M:%S'),
             num_new_links,
@@ -166,7 +166,8 @@ if __name__ == '__main__':
 
     # Step 1: Parse the links and dedupe them with existing archive
     links = merge_links(archive_path=out_dir, import_path=source, only_new=False)
-    
+    new_links = merge_links(archive_path=out_dir, import_path=source, only_new=True)
+
     # Step 2: Write new index
     write_links_index(out_dir=out_dir, links=links)
 
@@ -175,7 +176,6 @@ if __name__ == '__main__':
 
     # Step 4: Run the archive methods for each link
     if ONLY_NEW:
-        new_links = merge_links(archive_path=out_dir, import_path=source, only_new=True)
         update_archive(out_dir, new_links, source=source, resume=resume, append=True)
     else:
         update_archive(out_dir, links, source=source, resume=resume, append=True)
--- a/archiver/archive.py
+++ b/archiver/archive.py
@@ -10,7 +10,10 @@ from datetime import datetime
 from subprocess import run
 
 from parse import parse_links
-from links import validate_links
+from links import (
+    new_links,
+    validate_links
+)
 from archive_methods import archive_links, _RESULTS_TOTALS
 from index import (
     write_links_index,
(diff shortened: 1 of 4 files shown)


archiver/archive.py
```diff
@@ -158,13 +165,17 @@ def update_archive(archive_path, links, source=None, resume=None, append=True):
         source = download_url(source)
 
     # Step 1: Parse the links and dedupe them with existing archive
-    links = merge_links(archive_path=out_dir, import_path=source)
-    
+    links = merge_links(archive_path=out_dir, import_path=source, only_new=False)
+    new_links = merge_links(archive_path=out_dir, import_path=source, only_new=True)
+
     # Step 2: Write new index
     write_links_index(out_dir=out_dir, links=links)
```
pirate: I think you're missing `links=new_links` here, otherwise the index wont be updated to include the new links.

archiver/links.py
```diff
@@ -74,6 +74,21 @@ def validate_links(links):
...
+def new_links(imported_links, existing_links):
+    """
+    Return all links which are in the imported_links but not in the existing_links.
+    This is used to determine which links are new and not indexed jet. Set the
+    ONLY_NEW environment variable to activate this filter mechanism.
+    """
+    new_links = []
+    for i_link in imported_links:
+        found_link_in_existing_links = False
+        for e_link in existing_links:
```
pirate: This O(n^2) loop will crash for users who have many links (a few people have archived 100k+ links!).

Can you make a dictionary for `existing_links` that maps urls to link objects so this can be an O(n) loop.

archiver/archive.py
f0086: No, that is fine. `links` contain all links, including the new one. `new_links` only include the new links, so the newly generated index would contain only the new links.

archiver/links.py
f0086: Good point, I've updated the PR accordingly.

archiver/links.py
```diff
@@ -74,6 +74,14 @@ def validate_links(links):
...
     return list(links)
 
+def new_links(all_links, existing_links):
+    """
+    Return all links which are in the all_links but not in the existing_links.
+    This is used to determine which links are new and not indexed jet. Set the
+    ONLY_NEW environment variable to activate this filter mechanism.
+    """
+    existing_urls = list(map(lambda l: l['url'], existing_links))
+    return list(filter(lambda l: l['url'] not in existing_urls, all_links))
```
pirate: I think this is still O(n^2) because it'll iterate though all of `existing_urls` to find a match, instead of doing an O(1) lookup into a `dict`/`set` of urls.

Here's an O(n) solution that could work:

```suggestion
existing_urls = {l['url'] for l in existing_links}
return [link for link in all_links if link['url'] not in existing_urls]
```

f0086: Ok, I am not a Python programmer by any means. But I am sure my solution is O(2n), which is essentially O(n). I've actually benchmark this to see if the execution time is linear. Surprisingly, you are right, my solution IS in fact exponential, but if I remove the list() wrapper from the first line, it IS linear. I do not know the internals of map and filter in Python, but without the list() wrapper, 200.000 Bookmarks will take ~6 seconds on my machine (same as your solution).

So, map and filter runs in O(n) (Still not sure what magic the list() thing is doing internally). But nevertheless, I would like to use your solution, because it matches the coding style of the rest of the code.

PR #115, opened 2018-11-26
Fix 'Too many open files' error.

karlicoss:
Happened before after continuous archiving of few hundreds links (at 800-900th url typically). 

If you run archaver and in parallel, `lsof -p $(pgrep -f archive)`, you'll see PIPE file descriptors leaking.

Fixed by:
* setting process object to `None` to trigger GC finalizer cleanup of pipe descriptors
* protecting against double cleanup


pirate:
Wow nice analysis, thanks!

PR #116, opened 2018-11-27
Basic integration test fixture and couple of tests

karlicoss:
I'm investigating some issue with a bad state of the index, not sure if it's an actual bug or I just messed it up  somehow, but while investigating I started writing integration tests. It's a bit ad-hoc for now, but hopefully better to have at least some :) Perhaps later we can set them up on CI or something like that.


pirate:
Awesome work, thanks for adding this!

PR #117, opened 2018-12-01
README: fix donations link

FiloSottile:
BTW, I'd be happy to regularly kick in some money to support the project, maybe look into recurring donation options, Patreon being the most popular (and while it's not the cheapest, the network effect is valuable IMHO).


pirate:
I appreciate it! https://www.patreon.com/theSquashSH

I set one up a while ago but never got up the courage to share it around.

PR #118, opened 2018-12-01
Make table filterable, sortable and provide pagination

### LINKED ISSUE #40: Index: Add a way to sort the index table by column
Currently, archived items on the index page are sorted by reverse chronological order (most recent first), most likely they are presented in the same order as the original HTML bookmarks export.

It would be nice to be able to sort the table by clicking on each column's header. Eg.

 * Sort items by reverse/chronological order
 * Sort items by title
 * Sort items by original URL

This would probably require a bit of Javascript

### LINKED ISSUE #61: Index: Add a way to search the index titles, dates, and urls in the frontend

5hay:
fixes #61, fixes #40 

The table has advanced features now. One can filter over all columns, sort all
columns asc/desc and paginate the table.
The state (filtered result, current page, sorting order and number of
entries shown) will be saved in localStorage so the user can open the
page in another window or tab and see the same output again.

This is done with the help of https://github.com/DataTables/DataTables.

### DIFF
--- a/archiver/templates/index.html
+++ b/archiver/templates/index.html
@@ -38,11 +38,14 @@
             header h1 small a {
                 text-decoration: none;
                 color: orange;
-                opacity: 0.6
+                opacity: 0.6;
                 font-weight: 300;
             }
             header h1 small a:hover {
                 opacity: 1;
+            }
+            header + div {
+                margin-top: 10px;
             }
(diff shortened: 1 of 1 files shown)


pirate:
Wow this is awesome!  Thank you so much <3

PR #125, opened 2018-12-21
Fix the setup/archive symlinks

W1lkins:
Since the README still includes these in the instructions, I guess we should fix these symlinks.


pirate:
Can you also add the symlinks for the old name pointing to the new one so we don't break everyone's use of the old symlinks?

Thanks!

PR #147, opened 2019-02-19
Don't run ls if using non-default directory

nicolaschan:
Previously, an error message would appear if using a non-default directory because it would run `ls` on the non-existent default directory:

```bash
ls: cannot access '/home/nicolas/Library/Application Support/Firefox/Profiles/*.default/places.sqlite': No such file or directory
```

# Summary

This PR fixes an error message that appears when exporting browser history from a non-default directory (such as on Ubuntu, where Firefox data is at `~/.mozilla/firefox/`).

# Changes these areas

- [ ] Config
- [x] Bugfixes
- [ ] Command line interface
- [ ] Feature behavior
- [ ] Internal design


pirate:
Thanks so much! <3

PR #156, opened 2019-03-02
Fix double-download of Chromium binary and backwards symlink in Dockerfile

raman325:
# Summary

Added env variable to skip puppeteer's downloading of chromium because the Dockerfile already does that and not skipping it results in an error. Also removed a redundant symlinking step since the repo already includes those links

# Changes these areas

- [ ] Config
- [x] Bugfixes
- [ ] Command line interface
- [ ] Feature behavior
- [ ] Internal design

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ RUN git clone https://github.com/pirate/ArchiveBox /home/pptruser/app \
     && chown -R pptruser:pptruser /data \
     && ln -s /data /home/pptruser/app/archivebox/output \
     && ln -s /home/pptruser/app/bin/* /bin/ \
-#    && ln -s /home/pptruser/app/bin/archive /bin/archivebox \
+    # && ln -s /home/pptruser/app/bin/archive /bin/archivebox \
     && chown -R pptruser:pptruser /home/pptruser/app/archivebox
     # && pip3 install -r /home/pptruser/app/archivebox/requirements.txt
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,4 +69,4 @@ USER pptruser
 WORKDIR /home/pptruser/app
 
 ENTRYPOINT ["dumb-init", "--"]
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
             - SHOW_PROGRESS=False
         volumes:
             - ./data:/data
-        command: bash -c 'echo "https://github.com/pirate/ArchiveBox" | /bin/archive; tail -f /dev/null'
+        command: bash -c 'echo "https://github.com/pirate/ArchiveBox" | /bin/archivebox; tail -f /dev/null'
```
pirate: I'd rather use `/bin/archive` to stay consistent with the rest of the docs for ArchiveBox usage with the `./archive` shortcut in the project root.

Dockerfile
```diff
@@ -64,4 +69,4 @@ USER pptruser
 WORKDIR /home/pptruser/app
 
 ENTRYPOINT ["dumb-init", "--"]
-CMD ["/bin/archive"]
+CMD ["/bin/archivebox"]
```
pirate: Can you leave this as `/bin/archive` for now to stay consistent with the rest of the docs.

Dockerfile
```diff
@@ -45,7 +50,7 @@ RUN git clone https://github.com/pirate/ArchiveBox /home/pptruser/app \
     && chown -R pptruser:pptruser /data \
     && ln -s /data /home/pptruser/app/archivebox/output \
     && ln -s /home/pptruser/app/bin/* /bin/ \
-    && ln -s /home/pptruser/app/bin/archive /bin/archivebox \
```
pirate: Changed this line to the correct link (I had it backwards) instead of removing it:
`&& ln -s /home/pptruser/app/bin/archivebox /bin/archive \`

raman325:
Thanks for the clarification, your comments make much more sense 😄

PR #193, opened 2019-03-26
wget: Remove unsupported compression option

sbrl:
# Summary
`--compression=auto` is not supported on Raspberry Pis. This PR removes it from ArchiveBox.

See https://unix.stackexchange.com/questions/464372/how-could-compression-be-missing-from-my-wget for more information.

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/archive_methods.py
+++ b/archivebox/archive_methods.py
@@ -33,6 +33,7 @@ from config import (
     WGET_USER_AGENT,
     CHECK_SSL_VALIDITY,
     COOKIES_FILE,
+    WGET_AUTO_COMPRESSION
 )
 from util import (
     domain,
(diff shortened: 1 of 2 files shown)


pirate:
Can you change this to remove it based on `config.WGET_VERSION`, instead of outright on all devices even when it's supported.

sbrl:
@pirate I don't think so, unfortunately! The problem is that it depends on whether wget was _built_ with compression support, rather than just whether the version is recent enough (btw, does anyone know of a distro with wget 1.19+? Both Raspbian & Ubuntu seem to have an old version).

This test in bash would correctly identify if support is available:

```bash
compression_supported=0;
wget --compression=auto --help >/dev/null 2>&1;
if [[ "$?" -eq "0" ]]; then compression_supported=1; fi
```

Any idea how to do that in Python?

pirate:
```python
WGET_AUTO_DECOMPRESS = not run([WGET_BINARY, "--compression=auto", "--help"], stdout=DEVNULL).returncode
```

I'll just stick this in `config.py` in the next release.

sbrl:
I've updated this PR with the suggested changes - thanks! I'm unsure as to the conditional syntax in the `wget` options list though - I've tried to copy what's there already, but it looks a bit like magic to me.....

PR #194, opened 2019-03-26
util.py: Don't use regular expressions to parse HTML

sbrl:
# Summary
Regexes should never be used to parse HTML ([source](http://htmlparsing.com/regexes.html), and [another](https://stackoverflow.com/a/590789/1460422)).

This commit fixes the <title /> finding logic to use a full HTML parser instead.

Example URL that now works under the new system: https://css-tricks.com/snippets/css/complete-guide-grid/

Screenshot:
image

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/util.py
+++ b/archivebox/util.py
@@ -193,12 +193,14 @@ def fetch_page_title(url, timeout=10, progress=SHOW_PROGRESS):
         if progress:
             sys.stdout.write('.')
             sys.stdout.flush()
-
-        html = download_url(url, timeout=timeout)
-        document = lxml.html.parse(html)
         
-        result = document.find(".//title")
-        return result.text.strip() if result else None
+        html = download_url(url, timeout=timeout)
+        document = lxml.html.fromstring(html)
+        result = document.xpath(".//title")
+        if len(result) == 0:
+            return None
+        
+        return result[0].text.strip()
     except Exception as err:  # noqa
         # print('[!] Failed to fetch title because of {}: {}'.format(
         #     err.__class__.__name__,
--- a/archivebox/util.py
+++ b/archivebox/util.py
@@ -3,6 +3,7 @@ import re
 import sys
 import time
 
+import lxml.html
 from urllib.request import Request, urlopen
 from urllib.parse import urlparse, quote
 from decimal import Decimal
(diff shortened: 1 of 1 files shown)


pirate:
I'm surprised you didn't link to the famous SO post 😉 https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454

We use regex because it can handle broken HTML better than `parser = lxml.etree.XMLParser(recover=True)`.  The failure rate for titles is currently less than 1 in a hundred in my archive, and the ones that are failing are 404 pages, not dues to bad parsing, are you having a specific issue with certain titles?

sbrl:
Indeed, I'm having an issue in particular with [this url](https://css-tricks.com/snippets/css/complete-guide-grid/). Under the old system I get "mt" instead of the actual page title. I think it has multiple titles (for inline iframes or something?), so it's the _first_ one you want.

Apparently, [BeautifulSoup](https://stackoverflow.com/a/19548832/1460422) is capable for parsing very badly formatted HTML. What about if I extended this PR to fall-back on BeautifulSoup if lxml fails (a test URL that fails under lxml would be great if you know of one, but I'm sure I can write a program to find one if you don't have one to hand).

pirate:
All three examples you gave work fine with the regex we have right now, and a few of them would not be doable with BeautifulSoup.

```python
HTML_TITLE_REGEX = re.compile(
    r'<title.*?>'                      # start matching text after <title> tag
    r'(.[^<>]+)',                      # get everything up to these symbols
    re.IGNORECASE | re.MULTILINE | re.DOTALL | re.UNICODE,
)
```

## Example 1
```python
>>> example = """<title
... foo="bar">
...
... Yay
... <strong> I can break ArchiveBox :P</strong>
... </title>"""
>>> match = re.search(HTML_TITLE_REGEX, example)
>>> match.group(1).strip()
'Yay'
```
`Yay` is what's expected, tags inside of `<title>` are invalid, so we take everything up until the first closing tag.

## Example 2

```python
>>> example = """<title>I'm a title, but the author forgot to close the tag
...
... <script>
...
... yay
...
... ...."""
>>> match = re.search(HTML_TITLE_REGEX, example)
>>> match.group(1).strip()
"I'm a title, but the author forgot to close the tag"
```
Worked correctly, it stops at the first `<` so it can handle incorrect closing tags (which BeautifulSoup cant)

## Example 3
```python
>>> example = """<title>I'm a normal title</title>
...
... .....later on.....
... <title> I'm another title</title>"""
>>> match = re.search(HTML_TITLE_REGEX, example)
>>> match.group(1).strip()
"I'm a normal title"
```
It correctly gets only the first title.

## Given URL

When I tested it with the URL you specified it also worked fine:
```python
>>>
>>> from archivebox.util import fetch_page_title
>>> fetch_page_title('https://css-tricks.com/snippets/css/complete-guide-grid/')
'A Complete Guide to Grid | CSS-Tricks'
```

You can test with `fetch_page_title` on `dev` to get the latest version, but I'm pretty sure most of these work on master too.

sbrl:
Ah - my mistake! I missed the multiline regex flag there :P

Huh, that's weird. I just re-tested the problematic urls in the latest `master`, and they appear to work as intended now! I guess it was a different bug that was fixed elsewhere.

The original issue was that it was returning `mt` as the title instead of the actual title, but they appear to be functioning at intended now :D

pirate:
Glad to hear it :) Lemme know if you have other title parsing issues in the future and I'd be happy to revisit improving the parsing system.

sbrl:
Will do!

Thanks again :-)

PR #197, opened 2019-03-27
v0.3.0 (final release before moving to Django)

pirate:
# The v0.3.0 Release

A bunch of big changes:
 - entire codebase is now mypy type-checked, with runtime type enforcement so that a bug doesn't lead to malformed indexes being written to disk
 - all disk writes are now atomic so that power loss / SIGTERM during execution doesn't leave the index in an inconsistent state (as long as the archive folder is on a local, non-network filesystem)
 - core data model now uses typed, immutable dataclasses instead of plain `dict`s
 - schema and json loading and parsing are formalized all in one place `schema.py` (laying the foundations for a future transition to Django models)
 - got rid of all derivable redundant data in the index files
 - Must faster performance when passing over old links by avoiding excessive index writes
 - Entirely new UI for the Index Details page:

Screen Shot 2019-03-27 at 3 49 46 AM


noqqe:
cant wait to see this in master!  😍

pirate:
FYI for people following this PR, the next release after this PR is also up already: https://github.com/pirate/ArchiveBox/pull/207

The plan is to release both versions (nearly) simultaneously. Users that want to stay on the stable pre-django version can use v0.3.0, and users who want the latest cutting-edge release can install v0.4.0 and enjoy the new CLI experience.

pirate:
Ok I think this is done, I'm going to close this PR and recommend everyone upgrade straight to v0.4.0 when it's released (unless you have a specific reason to use this version).

PR #201, opened 2019-03-30
PEP-8 and steps towards testing

thesus:
# Summary
First of all thanks for creating and maintaining ArchiveBox! It's nice to know that some parts of the internet won't vanish.
I couldn't help myself but making most of the code PEP-8 conform for better readability and maintainability. Since pytest was used for testing at the beginning, python requirements were added to allow linting, coverage and testing with it. At the moment some parts are explicitly excluded because they required major changes. e.g. `archivebox/config.txt` My first thought was to rebase the dev branch to master but merging seemed to be simpler in this case.

I'll be by no means offended if you close this PR without further notice as it would've been wiser to communicate such changes beforehand. :see_no_evil:  I'd volunteer to add tests later if you're interested.

One can install the requirements via `pip install -r requirements/development.txt` and run linting and testing with `pytest`

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [x] Internal architecture


pirate:
Sorry man, but I am going to close this.  I really appreciate the effort though!

I have wide monitors and hate prematurely wrapped lines, makes it much harder to scan the code quickly imo.  .e.g stuff like this is way more confusing/bug-prone than just leaving it long:

image

In general, all the PEP's that I'm ignoring are explicitly PEPs that I disabled on my machine's flake8 checker.  They're not just things I've casually forgotten, they're intentionally disabled because I prefer the non-pep style.

Another example is `### ` comments vs `#` comments mean different things.  I use them like Markdown headers in reverse, so the more hash marks there are the more important the heading is. 

FYI `requirements.txt` is by default dev requirements, there's no need to create a separate file for base requirements as those are defined in `setup.py`.

If you like, you're welcome to reopen a separate PR with your fixes for the tests, I'll be adding more tests in the coming week myself as well.

PR #203, opened 2019-03-31
Add an option to skip links that are broken.

shakkhar:
# Summary
I think it is nice if ArchiveBox can be configured to not create index entry for non-existent pages. I don't particularly care how it's done, just that it's possible. This patch shows one approach. I didn't change the default behavior. 
 
# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [X] Configuration options
- [ ] Internal architecture


shakkhar:
I just realized that this patch doesn't handle non-HTTP errors, e. g. unresponsive host.

pirate:
It's ok, I'm doing this in a different way and will release it after testing on master shortly.

PR #216, opened 2019-04-12
Copy project into image instead of cloning to avoid stale images on Docker Hub

bltavares:
Docker `RUN` statements cache based on the text of the command executed,
not the content of what it does to the image.

Since the command was cloning the project, and the text didn't change,
building the image would not update the code if the image was already
cached. This lead to a stale Docker image distributed on Docker Hub.

This could also cause some confusion, as modified code would not show up
on the image during the build process.

This commit changes the build process to copy the content of the project
into the image. Whenever a file changes it will trigger a new updated
image.

**IMPORTANT: Do not submit PRs with only formatting / PEP8 / line length changes, I will close them with great prejudice.  The PEP8 checks I don't follow are intentional. PRs for minor bugfixes, typos, etc are fine.**

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,8 +45,8 @@ RUN groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
     && chown -R pptruser:pptruser /node_modules
 
 # Install the ArchiveBox repository and pip requirements
-RUN git clone https://github.com/pirate/ArchiveBox /home/pptruser/app \
-    && mkdir -p /data \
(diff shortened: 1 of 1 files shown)


pirate:
Ah thank you so much, I was wondering why that was happening sometimes.

I think someone told how this worked long ago and I somehow forgot, but it makes total sense.

PR #246, opened 2019-06-03
fix init error

gisforgirard:
archivebox init fails with "TypeError: __init__() missing 1 required positional argument: 'on_delete'" - this fixes it and init works

### DIFF
--- a/archivebox/core/models.py
+++ b/archivebox/core/models.py
@@ -72,7 +72,7 @@ class Snapshot(models.Model):
 class SnapshotResult(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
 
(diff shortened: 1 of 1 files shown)


archivebox/core/models.py
```diff
@@ -72,7 +72,7 @@ def link_dir(self):
 class SnapshotResult(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
 
-    snapshot = models.ForeignKey(Snapshot)
+    snapshot = models.ForeignKey(Snapshot,on_delete=models.CASCADE)
```
amiryal: Without getting into what the code change does:
```suggestion
    snapshot = models.ForeignKey(Snapshot, on_delete=models.CASCADE)
```

pirate:
Thanks!

PR #281, opened 2019-10-13
Fix a leaking file descriptor.

Julian:
# Summary

Handles a `ResourceWarning` on Py3, and generally not leaking FDs especially on non-CPython.

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/index.py
+++ b/archivebox/index.py
@@ -187,7 +187,7 @@ def patch_links_index(link, out_dir=OUTPUT_DIR):
     # Patch HTML index
     html_path = os.path.join(out_dir, 'index.html')
     with open(html_path, 'r') as html_file:
-        html = [line[:-1] for line in html_file]
+        html = html_file.read().splitlines()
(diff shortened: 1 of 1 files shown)


archivebox/index.py
```diff
@@ -186,7 +186,8 @@ def patch_links_index(link, out_dir=OUTPUT_DIR):
 
     # Patch HTML index
     html_path = os.path.join(out_dir, 'index.html')
-    html = open(html_path, 'r').read().split('\n')
+    with open(html_path, 'r') as html_file:
+        html = [line[:-1] for line in html_file]
```
pirate:
```suggestion
        html = html_file.read().splitlines()
```

Handles the case where the last line doesn't have a newline instead of chopping off the last character.

Julian:
(updated, thanks)

pirate:
Thanks!

PR #311, opened 2020-01-21
Add shebang to script file

### LINKED ISSUE #243: Configuration: Docker-compose exec fails with OCI runtime exec format error
I'm setting up ArchiveBox using the Docker image and instructions found [here](https://github.com/pirate/ArchiveBox/wiki/Docker#docker-compose). Though it appears to have been successfully setup on my computer, I cannot use the main command to add an individual link

```echo "https://example.com" | docker-compose exec -T archivebox /bin/archive```

I expected the outputted archive data to be stored in my data/ volume but it does not appear to be, and I get the following error ```OCI runtime exec failed: exec failed: container_linux.go:344: starting container process caused "exec format error": unknown```

eSoares:
Execution scripts without shebang cause cause "exec format error" in docker and elsewhere.
More info:
- https://www.lewuathe.com/exec-format-error-in-docker-container.html
- https://stackoverflow.com/a/12911604

Fixes #243

### DIFF
--- a/bin/archivebox
+++ b/bin/archivebox
@@ -1,1 +1,3 @@
+#!/bin/sh
+
 ../archivebox/archive.py


pirate:
This is a symlink though, not a normal file, you cant add new lines to it.

PR #322, opened 2020-02-15
guess encoding via chardet if available

michaelbub:
I had a page where the title would not be detected because in the headers there was no encoding info and by defaulting to utf-8 the html could not be decoded. The page itself says it is "content="text/html; charset=windows-1252".
By using chardet I could make it work. This was not tested at all except for the one page in question.


pirate:
thanks!

PR #357, opened 2020-07-07
Test Mock alternative. Proof of concept.

cdvv7788:
# Summary

There are cases where playing with the headers and other information in the urls we pass to `archivebox` is actually useful. If we use something like `httmock` we would need to know a lot about the internals of several bits of the application, and the tests would need to be updated so they read the mocked version instead of the original one.
Instead of that, we could run a small webserver that serves real content locally, and we can update the tests to consume the content from there. This better reflects the actual behavior the application will have in real usage, and the tests will be kept clean, as the application will be doing exactly the same it does outside testing.

This is a small POC that runs a `bottle` server while the tests are running.

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,6 +65,7 @@ setuptools.setup(
             "sphinx-rtd-theme",
             "recommonmark",
             "pytest",
+            "bottle",
         ],
         # 'redis': ['redis', 'django-redis'],
         # 'pywb': ['pywb', 'redis'],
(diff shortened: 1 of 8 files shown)


pirate:
yeah good idea, also I love bottle, good choice

cdvv7788:
@pirate I finished changing the tests to use the local webserver instead of remote requests. Please check.

PR #358, opened 2020-07-08
Django public views

cdvv7788:
# Summary

Currently, all of the views in django (outside of the admin) are public. I am not sure if that is intended.
I added the `LoginRequiredMixin` to all of them.
If they are meant to be public, just close the PR.

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/core/views.py
+++ b/archivebox/core/views.py
@@ -5,6 +5,7 @@ from django.shortcuts import render, redirect
 from django.http import HttpResponse
 from django.views import View, static
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


pirate:
it's supposed to use the config to set the permissions:

image

cdvv7788:
I will close it for now. However I think that some mixins can be used to ensure that everything is secured as expected. There are checks per view per method in some cases, but there are some unchecked methods. For example, in `add`, the `post` method has no check, so it is basically public.

PR #361, opened 2020-07-09
fix: Create user home to avoid warning

cdvv7788:
# Summary

When trying to run the container, I was getting the following error:
```
mkdir: cannot create directory ‘/home/archivebox’: Permission denied
touch: cannot touch '/home/archivebox/.local/share/applications/mimeapps.list': No such file or directory
```

I added the flag to create a home directory during the build step, so this warning does not show up.

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


apkallum:
Looks good & tested it!

PR #365, opened 2020-07-16
fix: htmlencode titles before rendering the static html index and detail

cdvv7788:
# Summary

The content is being decoded after it is read. This makes sense in some areas, and it can be a much deeper conversation. However, when displaying static html (index.html, link_details, etc) it should be encoded back. Also note that, when decoding, there may be some information loss, as the decoding -> encoding can return a different output than the original.

This PR fixes the situation where a decoded html tag is rendered in the static index and detail pages, effectively breaking the layout. This only affects the static files. The other indexes are left untouched.

**Related issues: https://github.com/pirate/ArchiveBox/issues/330**

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/index/html.py
+++ b/archivebox/index/html.py
@@ -90,7 +90,7 @@ def main_index_row_template(link: Link) -> str:
         **link._asdict(extended=True),
         
         # before pages are finished archiving, show loading msg instead of title
-        'title': (
+        'title': htmlencode(
             link.title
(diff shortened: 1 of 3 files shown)


apkallum:
Looks good! I remember seeing this issue. Thanks Cristian.

PR #373, opened 2020-07-20
 fix: Remove imports causing circular import issues

cdvv7788:
# Summary

The logging module is causing issues because it is importing things outside its scope. Most of the problem is the typing. I removed that for now to make everything work again. We can discuss a better solution later.

**Related issues: https://github.com/pirate/ArchiveBox/issues/372

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/logging.py
+++ b/archivebox/logging.py
@@ -410,7 +410,7 @@ def log_removal_finished(all_links: int, to_keep: int):
 
 
 def log_shell_welcome_msg():
-    from . import list_subcommands
+    from .cli import list_subcommands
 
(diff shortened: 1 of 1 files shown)


pirate:
Instead of removing these imports, you should change it to:
```python3
from typing import TYPE_CHECKING

if TYPE_CHECKING:
   import ...
```

cdvv7788:
@pirate check again. This seems to work.

Issue #257, opened 2019-08-16
Error if the character code is Shift_JIS


matoken:
#### Error if the character code is Shift_JIS

An error occurs on some Japanese pages.

#### Steps to reproduce

An error occurs on this page.

https://www.mbc.co.jp/news/
http://167.86.112.42/hello_sjis.html

#### Screenshots or log output

```
$ ./archive https://www.mbc.co.jp/news/
[*] [2019-08-16 18:53:59] Downloading https://www.mbc.co.jp/news/
[!] Failed to download https://www.mbc.co.jp/news/

	 'utf-8' codec can't decode byte 0x8e in position 181: invalid start byte
```

The character code of the page where the error occurs seems to be Shift_JIS (a little old Japanese character code).
```
$ curl -s https://www.mbc.co.jp/news/ | grep -i charset=
<meta http-equiv="content-type" content="text/html; charset=Shift_JIS" />
<script type="text/javascript" src="js/scrollsmoothly.js" charset="utf-8"></script>
<link rel="stylesheet" type="text/css" href="/css/mbc_menu_import.css" charset="Shift-JIS">
<SCRIPT language="JavaScript" src="/js/mbcmenu.js" charset="Shift-JIS"></SCRIPT>
```
An error occurred when trying to create a tiny Shift_JIS page.

```
$ echo '<html>
<head>
<meta http-equiv="content-type" content="text/html; charset=Shift_JIS" />
</head>
<body>
こんにちは
</body>
</html>' | iconv -f UTF8 -t SJIS > hello_sjis.html
```

```
$ ./archive http://167.86.112.42/hello_sjis.html
[*] [2019-08-16 19:02:10] Downloading http://167.86.112.42/hello_sjis.html
[!] Failed to download http://167.86.112.42/hello_sjis.html                                                                                                                      

     'utf-8' codec can't decode byte 0x82 in position 103: invalid start byte
```

#### Software versions

 - OS: Debian GNU/Linux 10 (buster) amd64
 - ArchiveBox version: ArchiveBox version e2b054ae7
 - Python version: 3.7.3 ( Debian package 3.7.3-1 )
 - Chrome version: 73.0.3683.75 (Debian package 73.0.3683.75-1 )

pirate:
[...]

PR #378, opened 2020-07-22
fix: Use w3lib to improve the encoding extraction

cdvv7788:
# Summary

Detecting the right encoding is an issue not only for rss feeds (which had a previous fix) but in general. In this PR I generalize that fix, so the flow is:
- Headers
- Body
- Guess it (requests default behavior)

image

Problematic links are all having the right title now.

**Related issues: #257

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setuptools.setup(
         "ipython",
         "youtube-dl",
         "python-crontab==2.5.1",
-        "w3lib==v1.22.0",
+        "w3lib==1.22.0",
         # "croniter",
         # Some/all of these will likely be added in the future:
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -47,6 +47,7 @@
         "ipython",
         "youtube-dl",
         "python-crontab==2.5.1",
+        "w3lib==v1.22.0",
```
pirate, line 50:
```suggestion
        "w3lib==1.22.0",
```

Issue #374, opened 2020-07-20
Bugfix: django branch start_ts error on init


drpfenderson:
#### Describe the bug
When attempting to `archivebox init` with version 0.4.3 in old archive, archivebox fails at `Collecting links from any existing indexes and archive folders...` with `KeyError: 'start_ts'`

#### Steps to reproduce
1. Installed Django branch with `git clone` and `pip install .`.
2. Navigated to old archive directory.
3. Ran `archivebox init`
4. archivebox goes through most of importing process, and then dies with the error listed below.

#### Screenshots or log output

```
Traceback (most recent call last):
  File "/home/USERNAME/.local/bin/archivebox", line 8, in <module>
    sys.exit(main())
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/cli/__init__.py", line 126, in main
    pwd=pwd or OUTPUT_DIR,
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/cli/__init__.py", line 62, in run_subcommand
    module.main(args=subcommand_args, stdin=stdin, pwd=pwd)    # type: ignore
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/cli/archivebox_init.py", line 34, in main
    out_dir=pwd or OUTPUT_DIR,
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/util.py", line 108, in typechecked_function
    return func(*args, **kwargs)
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/main.py", line 316, in init
    for link in load_main_index(out_dir=out_dir, warn=False)
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/util.py", line 108, in typechecked_function
    return func(*args, **kwargs)
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/index/__init__.py", line 250, in load_main_index
    all_links = list(parse_json_main_index(out_dir))
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/index/json.py", line 52, in parse_json_main_index
    yield Link.from_json(link_json)
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/index/schema.py", line 203, in from_json
    cast_result = ArchiveResult.from_json(json_result)
  File "/home/USERNAME/.local/lib/python3.7/site-packages/archivebox/index/schema.py", line 62, in from_json
[...]

PR #380, opened 2020-07-22
fix: Ignore incomplete index entry, and complete partially complete ones

cdvv7788:
# Summary

The init process tries to parse some data from alternative sources instead of crashing. This should work with old indexes.
Incomplete entries (the ones without `cmd`) are ignored and removed from the index.

**Related issues: #374 

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/index/schema.py
+++ b/archivebox/index/schema.py
@@ -2,6 +2,7 @@ __package__ = 'archivebox.index'
 
 import os
 
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


cdvv7788:
Wrong approach

PR #388, opened 2020-07-23
fix: Remove link from sql index on remove command

cdvv7788:
# Summary

When using `archivebox remove link`, it was not being removed from the sql index.

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
nice test too 👍

PR #389, opened 2020-07-24
fix: Guess timestamps and add placeholders to support older indices

cdvv7788:
# Summary

Index loading is more liberal now. Instead of failing, it will attempt to repair the found files by extracting the timestamp from the `duration` and `timestamp` fields. Other values are just marked as `undefined`.
This should not affect up to date indexes. Old indexes, however, should get rewritten with the new information.

**Related issues: #374 

# Changes these areas

- [X] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/index/schema.py
+++ b/archivebox/index/schema.py
@@ -1,6 +1,7 @@
 __package__ = 'archivebox.index'
 
 import os
+from pathlib import Path
 
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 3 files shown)


archivebox/index/schema.py
```diff
@@ -51,17 +52,39 @@ def typecheck(self) -> None:
...
-        info['cmd_version'] = info.get('cmd_version')
+        if guess:
+            keys = info.keys()
+            if "start_ts" not in keys:
+                info["start_ts"], info["end_ts"] = cls.guess_ts(json_info)
+            else:
+                info['start_ts'] = parse_date(info['start_ts'])
+                info['end_ts'] = parse_date(info['end_ts'])
+            if "pwd" not in keys:
+                info["pwd"] = str(os.getcwd() / Path(f"archive/{json_info['timestamp']}"))
```
pirate, line 79: OUTPUT_PATH + ARCHIVE_DIR_NAME instead of cwd + `archive/`

cdvv7788: Good catch

Issue #409, opened 2020-07-31
Bug: 500 Error after finishing archive request.


PokeyTheCat:
Using docker and `ArchiveBox v0.4.9` web server interface. I'm getting a `Server Error (500)` when the post completes after I submit a new archive request in the web.

This is a fresh install and newly initialized instance. Can I increase the verbosity to see what's making it mad?

```
REPOSITORY                TAG                 IMAGE ID            CREATED             SIZE
nikisweeting/archivebox   latest              3e02bd28e479        5 hours ago         1.07GB
```

```
Attaching to archivebox_1
archivebox_1  | [i] [2020-07-30 22:44:34] ArchiveBox v0.4.9: archivebox server 0.0.0.0:8000
archivebox_1  |     > /data
archivebox_1  |
archivebox_1  | [+] Starting ArchiveBox webserver...
archivebox_1  | [i] The admin username is: archivebox
archivebox_1  | Performing system checks...
archivebox_1  |
archivebox_1  | System check identified no issues (0 silenced).
archivebox_1  | July 30, 2020 - 22:44:35
archivebox_1  | Django version 3.0.8, using settings 'core.settings'
archivebox_1  | Starting development server at http://0.0.0.0:8000/
archivebox_1  | Quit the server with CONTROL-C.
archivebox_1  | [30/Jul/2020 22:45:01] "POST /admin/login/?next=/admin/core/snapshot/add/ HTTP/1.1" 302 0
archivebox_1  | [30/Jul/2020 22:45:01] "GET /admin/core/snapshot/add/ HTTP/1.1" 200 8756
archivebox_1  | [30/Jul/2020 22:45:01] "GET /static/admin/css/dashboard.css HTTP/1.1" 200 412
archivebox_1  | [+] Adding URL: https://medium.com/@nielssj/docker-volumes-and-file-system-permissions-772c1aee23ca
archivebox_1  | [30/Jul/2020 22:45:19] "GET /static/admin/fonts/Roboto-Bold-webfont.woff HTTP/1.1" 200 86184
archivebox_1  | [+] [2020-07-30 22:45:19] Adding 1 links to index (crawl depth=0)...
archivebox_1  |     > Saved verbatim input to sources/1596149119-import.txt
archivebox_1  |     > Parsed 1 URLs from input (Plain Text)
archivebox_1  |     > Found 1 new URLs not already in index
archivebox_1  |
[...]

PR #410, opened 2020-07-31
#409: Change import that was not working

cdvv7788:
# Summary

Top level import not working as expected. Changed it to reference the main package (`archivebox.config`).

**Related issues: #409 

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/util.py
+++ b/archivebox/util.py
@@ -1,3 +1,5 @@
+__package__ = 'archivebox'
+
 import re
 import json as pyjson
 


archivebox/util.py
```diff
@@ -175,7 +175,7 @@ def download_url(url: str, timeout: int=None) -> str:
 def chrome_args(**options) -> List[str]:
     """helper to build up a chrome shell command with arguments"""
 
-    from .config import CHROME_OPTIONS
+    from archivebox.config import CHROME_OPTIONS
```
pirate, line 178: hmm I don't think this is the right solution, I think you should put `__package__ = 'archivebox'` at the top of the file instead

Issue #69, opened 2018-03-10
Archive Method: Markdown output by using full-text extraction provided by a "reader mode" library


swhib:
Your program is great. I think it would be even better if it also saved a readability version which only tries to keep the relevant parts (reader view, clean view). I see two advantages

- full text searches should be more accurate and contain less irrelevant matches. I save a lot of interesting articles and usually they are surrounded by headings/links to other articles which will show up in my searches.
- It makes reading an article later much nicer. At the moment in Chromium 63 I can't use the most popular readability extension named Mercury Reader to transform a site form my disk that I have saved with bookmark-archiver. It's the same for Firefox 58.0.2. So I have to copy the actual content into a text editor or word processor to have a distraction-free reading view.

pirate:
This is already in progress, I'm working on adding a new archive method based on `mediumexporter` that extracts article body as markdown.  Once it's in markdown format any reader view can be used to view it.

pirate:
Debating between these options:

 - https://github.com/domchristie/turndown
 - https://github.com/gaojiuli/tomd
 - https://github.com/mozilla/readability

swhib:
Thank you very much for considering this issue. Here are some ideas (note that I'm not a programmer):

Is readability comparable to tomd or turndown? The first one cleans and the latter ones only convert? At least when I put a full site into the [turndown online converter](http://domchristie.github.io/turndown/) the result is only gibberish. So if you really wanted markdown woulnd't you have to chain these tools?

But I wonder if you really want to only offer markdown: Bookmark-archiver results are mainly viewed in the browser so there should be an option to view in cleaned in the browser, too? If you need html why convert to markdown and convert it back? I regularly convert articles to markdown for my personal offline evernote alternative. So I can see why an additional markdown export/option would be very useful.

For converting from html to markdown I personally prefer pandoc. Again my preference for pandoc is motivated by generel considerations: By far it's the most widely used converter, has the most contributors, etc. Also, Turndown seems to require [hundreds of dependencies](https://github.com/domchristie/turndown/blob/master/package-lock.json) whereas with pandoc you only have to trust one source (though this is a question of personal preference). I know that mozilla's readability also requires quite a few dependencies but a) much fewer and b) I hope that mozilla checks the sources of the dependencies (because it should have the manpower). I think requiring pandoc is no problem: There are binaries for every plattform. Putting such a binary in your PATH is easier than downloading your software from the github repo and running it.

[...]

PR #426, opened 2020-08-07
feat: Initial version of readability extractor

cdvv7788:
# Summary

Initial working version of the readability extractor, using https://github.com/cdvv7788/readability-extractor as wrapper (JSDOM, readability, dompurify).
This has a long way to go, but for now it is functional.

Questions:
- Currently, it accepts the html directly as an arg. When something fails, it will output the whole command, which will have a HUGE html attached to it...is that acceptable? Is it better to save it to a tmp_file and load the contents from there?
- The wrapper is pretty basic. It does what it needs at the bare minimum to work. Improvements there will definitely be necessary.
- Currently, the output is saved in a file named `readability.json` inside of the archive. It has several fields. What is the output you are expecting? Is it ok to keep this json or should we extract an specific field?

**Related issues: #69 

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

# Pending work

- [x] Add tests
- [x] Add `readability.json` to the detail html index
- [x] Add support for readability to the `Dockerfile`
- [x] Add support for readability to the CI pipeline


archivebox/extractors/readability.py
```diff
@@ -0,0 +1,83 @@
...
+@enforce_types
+def save_readability(link: Link, out_dir: Optional[str]=None, timeout: int=TIMEOUT) -> ArchiveResult:
+    """download reader friendly version using @mozilla/readability"""
+
+    out_dir = out_dir or link.link_dir
+    output = str(Path(out_dir).absolute() / "readability.json")
+
+    document = download_url(link.url)
+
+    # SingleFile CLI Docs: https://github.com/gildas-lormeau/SingleFile/tree/master/cli
```
pirate, line 42: fix docs url

archivebox/extractors/readability.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@enforce_types
+def save_readability(link: Link, out_dir: Optional[str]=None, timeout: int=TIMEOUT) -> ArchiveResult:
+    """download reader friendly version using @mozilla/readability"""
+
+    out_dir = out_dir or link.link_dir
+    output = str(Path(out_dir).absolute() / "readability.json")
+
+    document = download_url(link.url)
```
pirate, line 40: check for existing download from wget/singlefile/etc instead of downloading it from scratch

cdvv7788: I can do that, but I will leave the `download_url` call as a fallback (in case that, for example, only this method is being used)

archivebox/extractors/readability.py
```diff
@@ -0,0 +1,89 @@
...
+    READABILITY_BINARY,
+    READABILITY_VERSION,
+    CHROME_BINARY,
+)
+from ..logging_util import TimedProgress
+
+
+@enforce_types
+def should_save_readability(link: Link, out_dir: Optional[str]=None) -> bool:
+    out_dir = out_dir or link.link_dir
```
pirate, line 28:
```suggestion
    out_dir = out_dir or link.link_dir
    if is_static_file(link.url):
        return False
```

archivebox/extractors/readability.py
```diff
@@ -0,0 +1,89 @@
...
+
+from typing import Optional
+import json
+
+from ..index.schema import Link, ArchiveResult, ArchiveError
+from ..system import run, atomic_write
+from ..util import (
+    enforce_types,
+    download_url,
+
```
pirate, line 14:
```suggestion
    is_static_file,
```

cdvv7788:
The extractor is functional. There are a couple caveats that may need to be worked on:
- readability-extractor should have it's version pinned to avoid surprises
- A temp file is being created and not deleted. The OS should take care of it eventually. This extractor does not attempt to clean it (it could, if it is deemed necessary)

archivebox/extractors/readability.py
```diff
@@ -0,0 +1,113 @@
...
+@enforce_types
+def get_html(link: Link, path: Path) -> str:
+    """
+    Try to find wget, singlefile and then dom files.
+    If none is found, download the url again.
+    """
+    canonical = link.canonical_outputs()
+    abs_path = path.absolute()
+    sources = [canonical["wget_path"], canonical["singlefile_path"], canonical["dom_path"]]
```
pirate, line 33:
```suggestion
    sources = [canonical["singlefile_path"], canonical["wget_path"], canonical["dom_path"]]
```

PR #430, opened 2020-08-08
Archivebox not working when using only singlefile extractor

cdvv7788:
# Summary

https://github.com/pirate/ArchiveBox/commit/19aa5c3e94118c495a21ea291d56608cc96ccf47 was missing an update to `USE_CHROME`, so the flag was being set to `False` which in turn ended up changing `SAVE_SINGLEFILE` to `False` too.
This is an edge case, as it only happened when no `dom`, `picture` or `pdf` was being saved.

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
ah good catch, thanks

PR #451, opened 2020-08-18
Add options to ease management of node related extractors

cdvv7788:
**IMPORTANT: Do not submit PRs with only formatting / PEP8 / line length changes, I will close them with great prejudice.  The PEP8 checks I don't follow are intentional. PRs for minor bugfixes, typos, etc are fine.**

# Summary
Added new config, updated readme, and added an npm command to make it easier for users that want to install on bare metal.

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [X] Configuration options
- [ ] Internal architecture


README.md
```diff
@@ -2,12 +2,12 @@
...
-<a href="https://github.com/pirate/ArchiveBox">Github</a> | 
-<a href="https://github.com/pirate/ArchiveBox/wiki">Documentation</a> | 
-<a href="#background--motivation">Info & Motivation</a> | 
-<a href="https://github.com/pirate/ArchiveBox/wiki/Web-Archiving-Community">Community</a> | 
+▶️ <a href="https://github.com/pirate/ArchiveBox/wiki/Quickstart">Quickstart</a> |
+<a href="https://archivebox.zervice.io/">Demo</a> |
+<a href="https://github.com/pirate/ArchiveBox">Github</a> |
+<a href="https://github.com/pirate/ArchiveBox/wiki">Documentation</a> |
+<a href="#background--motivation">Info & Motivation</a> |
+<a href="https://github.com/pirate/ArchiveBox/wiki/Web-Archiving-Community">Community</a> |
```
pirate, line 10: careful, this whitespace is significant in markdown

README.md
```diff
@@ -82,9 +83,16 @@ open http://127.0.0.1:8000
 ```bash
 # Bare Metal
 # Use apt on Ubuntu/Debian, brew on mac, or pkg on BSD
+# You may need to add a ppa with a more recent version of nodejs
 apt install python3 python3-pip git curl wget youtube-dl chromium-browser
 
+curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
+  && echo 'deb https://deb.nodesource.com/node_14.x buster main' >> /etc/apt/sources.list \
```
pirate, line 90:
```suggestion
  && echo 'deb https://deb.nodesource.com/node_14.x $(lsb_release -cs) main' >> /etc/apt/sources.list \
```

package.json
```diff
@@ -5,7 +5,8 @@
 	"author": "Nick Sweeting <archivebox-npm@sweeting.me>",
 	"license": "MIT",
 	"scripts": {
-		"archivebox": "./bin/archive"
+   "setup": "node -e \"const {execSync} = require('child_process'); Object.entries(JSON.parse(fs.readFileSync('package.json')).dependencies).forEach(globaldep => execSync('npm i -g ' + globaldep[1]))\"",
```
pirate, line 8: I would remove this setup method, if they're viewing this `package.json` file then they're already in a repo dir where they can just run `npm install` rather than installing it system-wide. (as docker and github actions do)

README.md
```diff
@@ -82,9 +83,16 @@ open http://127.0.0.1:8000
...
+  && apt-get update -qq \
+  && apt-get install -qq -y --no-install-recommends nodejs
+
 pip install archivebox      # install archivebox
+npm run setup
```
pirate, line 95:
```suggestion
npm install -g 'git+https://github.com/pirate/ArchiveBox.git'
```

README.md
cdvv7788: argh, vscode messing with it. I will revert it.

package.json
cdvv7788: Sorry, forgot that the users normally don't clone the repository. Still, system wide should be better for users. I started using the docker version when readability and singlefile were mandatory because it was a pain to set the variables every time. If we ask the user to install them locally, we will need to add instructions to set `SINGLEFILE_BINARY` and `READABILITY_BINARY` as part of the installation, which just feels weird. Another option (I asked somewhere else I think) is to setup the binary by default to `node_modules/.bin/readability-extractor`.

README.md
cdvv7788: > careful, this whitespace is significant in markdown

Is this for the rendering in some other place than github? https://github.com/pirate/ArchiveBox/tree/node_config looks the same: https://github.com/pirate/ArchiveBox
I can revert it, just wondering what it breaks

PR #452, opened 2020-08-18
Replace index.json with index.sql as the main index

cdvv7788:
# Summary

After this PR is ready, the json index will not be considered the main source of truth anymore. Instead, the index.sqlite3 will replace it in that role.
The index.json will still be around, but it will only be written at the end of the processes that run. If the archive is old (no index.sqlite3 is present) running `archivebox init --force` will be necessary to update it to the latest version.

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [X] Internal architecture

# Roadmap Goals

This is one of the main goals of the 0.5 release.


archivebox/index/__init__.py
```diff
@@ -344,46 +325,6 @@ def dedupe_links(existing_links: List[Link],
 
     return all_links, new_links
 
-
-@enforce_types
-def patch_main_index(link: Link, out_dir: str=OUTPUT_DIR) -> None:
```
pirate, line 349: 🎉 removing this function is cause for celebration!

archivebox/cli/archivebox_list.py
```diff
@@ -46,6 +46,16 @@ def main(args: Optional[List[str]]=None, stdin: Optional[IO]=None, pwd: Optional
...
     )
+    group.add_argument(
+        '--html',
+        action='store_true',
+        help="Print the output in HTML format"
+    )
+    parser.add_argument(
+        '--index',
+        action='store_true',
+        help='Include the index additional structures'
```
pirate, line 57: This might be worth clarifying somewhere, what exactly are the additional structures?

cdvv7788: Maybe these are confusing...what about `--header` or `--include-header`? That is more explicit, and works for both html and csv.

pirate: I like `--headers` or `--with-headers`.

archivebox/extractors/__init__.py
```diff
@@ -136,24 +128,34 @@ def archive_link(link: Link, overwrite: bool=False, methods: Optional[Iterable[s
 
     return link
 
-
 @enforce_types
-def archive_links(links: List[Link], overwrite: bool=False, methods: Optional[Iterable[str]]=None, out_dir: Optional[str]=None) -> List[Link]:
-    if not links:
+def archive_links(all_links: any, overwrite: bool=False, methods: Optional[Iterable[str]]=None, out_dir: Optional[str]=None) -> List[Link]:
+
+    if type(all_links) is list:
```
pirate, line 134: I tend to prefer `isinstance` over `type(...) is ...`: https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance

In this case, however, I'd explicitly check for the QuerySet case, and fallback to generic python iteration otherwise, that way it still supports generators, tuples, etc.

archivebox/extractors/__init__.py
```diff
@@ -136,24 +128,34 @@ def archive_link(link: Link, overwrite: bool=False, methods: Optional[Iterable[s
 
     return link
 
-
 @enforce_types
-def archive_links(links: List[Link], overwrite: bool=False, methods: Optional[Iterable[str]]=None, out_dir: Optional[str]=None) -> List[Link]:
-    if not links:
+def archive_links(all_links: any, overwrite: bool=False, methods: Optional[Iterable[str]]=None, out_dir: Optional[str]=None) -> List[Link]:
```
pirate, line 132:
```suggestion
def archive_links(all_links: Union[Iterable[Link], QuerySet[Snapshot]], overwrite: bool=False, methods: Optional[Iterable[str]]=None, out_dir: Optional[str]=None) -> List[Link]:
```

archivebox/index/__init__.py
```diff
@@ -261,38 +246,30 @@ def write_main_index(links: List[Link], out_dir: str=OUTPUT_DIR, finished: bool=
...
 
+@enforce_types
+def write_static_index(links: List[Link], out_dir: str=OUTPUT_DIR) -> None:
+    with timed_index_update(os.path.join(out_dir, JSON_INDEX_FILENAME)):
+        write_json_main_index(links)
+    with timed_index_update(os.path.join(out_dir, HTML_INDEX_FILENAME)):
+        write_html_main_index(links, out_dir=out_dir, finished=True)
+
+@enforce_types
+def get_empty_snapshot_queryset(out_dir: str=OUTPUT_DIR):
```
pirate, line 257: Why do we need a function for this? Isn't `Snapshot.objects.none()` clearer on its own?

archivebox/index/__init__.py
```diff
@@ -322,67 +299,47 @@ def parse_links_from_source(source_path: str, root_url: Optional[str]=None) -> T
...
-    all_links = validate_links(existing_links + new_links)
-    all_link_urls = {link.url for link in existing_links}
-
-    new_links = [
-        link for link in new_links
-        if link.url not in all_link_urls
-    ]
+def fix_duplicate_links_in_index(snapshots: QuerySet, links: Iterable[Link]) -> Iterable[Link]:
+    """
+    Look for urls in the index, and merge them too
```
pirate, line 305:
```suggestion
    Given a list of in-memory Links, dedupe and merge them with any conflicting Snapshots in the DB.
```

archivebox/extractors/__init__.py
```diff
@@ -136,24 +128,34 @@ def archive_link(link: Link, overwrite: bool=False, methods: Optional[Iterable[s
...
+
+    if type(all_links) is list:
+        num_links: int = len(all_links)
+        get_link = lambda x: x
+        get_iter = lambda x: x
+    else:
+        num_links: int = all_links.count()
+        get_link = lambda x: x.as_link()
+        get_iter = lambda x: x.iterator()
```
pirate, line 141:
```suggestion
    if isinstance(all_links, QuerySet):
        all_links = all_links.iterator()
        num_links = all_links.count()
    else:
        num_links = len(all_links)
```

archivebox/index/html.py
```diff
@@ -31,6 +31,7 @@
 
 join = lambda *paths: os.path.join(*paths)
 MAIN_INDEX_TEMPLATE = join(TEMPLATES_DIR, 'main_index.html')
+MINIMAL_INDEX_TEMPLATE = join(TEMPLATES_DIR, 'main_index_minimal.html')
```
pirate, line 34: Why can't the minimal template just be a subset of the main template?

e.g.

- `main_index.html`
   ```html
   <html>
       <head>
           ...
       </head>
       <body>
           <header>...</header>
           <main>
               {% include 'main_index_table.html' %}
           </main>
           <footer>...</footer>
       </body>
- `main_index_table.html`

archivebox/index/sql.py
```diff
@@ -21,29 +22,35 @@ def parse_sql_main_index(out_dir: str=OUTPUT_DIR) -> Iterator[Link]:
...
-            Snapshot.objects.filter(url=link.url).delete()
+        snapshots.delete()
+
+@enforce_types
+def write_link_to_sql_index(link: Link):
+    from core.models import Snapshot
+    info = {k: v for k, v in link._asdict().items() if k in Snapshot.keys}
+    try:
+        info["timestamp"] = Snapshot.objects.get(url=link.url).timestamp
+    except Snapshot.DoesNotExist:
```
pirate, line 38: why does it increment int he case where it *doesn't* exist? shouldn't it increment in the case where it *does* already exist?

archivebox/main.py
```diff
@@ -252,7 +254,8 @@ def init(force: bool=False, out_dir: str=OUTPUT_DIR) -> None:
     """Initialize a new ArchiveBox collection in the current directory"""
     os.makedirs(out_dir, exist_ok=True)
     is_empty = not len(set(os.listdir(out_dir)) - ALLOWED_IN_OUTPUT_DIR)
-    existing_index = os.path.exists(os.path.join(out_dir, JSON_INDEX_FILENAME))
+
+    existing_index = (Path(out_dir) / SQL_INDEX_FILENAME).exists()
```
pirate, line 258: Can you add a check here for `JSON_INDEX_FILENAME`, it it exists show a warning saying it will no longer be kept up-to-date automatically, and that they should run `archivebox list --json > ...` manually when they want to use this file.

archivebox/main.py
```diff
@@ -317,13 +320,12 @@ def init(force: bool=False, out_dir: str=OUTPUT_DIR) -> None:
     print()
     print('{green}[*] Collecting links from any existing indexes and archive folders...{reset}'.format(**ANSI))
 
-    all_links: Dict[str, Link] = {}
+    all_links = get_empty_snapshot_queryset()
```
pirate, line 323:
```suggestion
    all_links = Snapshot.objects.none()
```

archivebox/main.py
```diff
@@ -597,7 +578,7 @@ def remove(filter_str: Optional[str]=None,
     
     check_data_folder(out_dir=out_dir)
 
-    if links is None:
+    if not snapshots:
```
pirate, line 581:
```suggestion
    if snapshots is None:
```

An empty QuerySet is still a valid QuerySet, only `None` is invalid.

pirate (changes requested):
From a high level this looks good, but one I think I was hoping is that we could get rid of `Link` entirely with this PR and use Snapshot.

Link was only needed as an in-memory representation of a Snapshot before it was added to the DB, because the DB was not the single source of truth.

Now that the DB is a single source of truth, we can do everything with `Snapshot`, and if we need an in-memory representation we can just do `Snapshot(url='http://...', ...)` without calling `.save()` yet.

archivebox/index/sql.py
cdvv7788: There was a case (the one that was causing issues for users) where the url didn't exist, but the assigned timestamp was one that was already assigned to something else, causing a collision when it was assigned. In this case, if the url is not created yet, the following block tries to find (in a naive way) a suitable timestamp that does not cause a collision in the index so the object can be created.

cdvv7788:
Thanks for the review! I will make the required changes, and start the removal of `Link` from archivebox.

archivebox/index/sql.py
pirate: ah, got it

archivebox/index/html.py
cdvv7788: These templates don't use django's rendering system. We should change them to use it, but that could be done in a later version.

archivebox/cli/archivebox_list.py
cdvv7788: https://github.com/pirate/ArchiveBox/pull/452/commits/ae94dbfd80b195ab74a0e0690a7d1e8de9665777

archivebox/extractors/__init__.py
cdvv7788: Depends...In this case I don't think it matters much. I changed the check: https://github.com/pirate/ArchiveBox/pull/452/commits/ec48f46f2deecbafb7dc0ed360b9e71a33efc9f5

archivebox/extractors/__init__.py
cdvv7788: How can I load? Importing Snapshot breaks because django has not been initialized. Unless I manually initialize it causes issues, and I am not a fan of that either, because I would need to make sure that wherever this function is called, django is initialized.

archivebox/index/__init__.py
cdvv7788: `RuntimeError: Model class archivebox.core.models.Snapshot doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`. I need to re-run setup_django with check_db=True (which is what I do in that function).
Once django is initialized before anything else, we could change it with no issues.

archivebox/extractors/__init__.py
cdvv7788: I need to differentiate between a `Snapshot` and a `Link`. If I do this, I would have to check the type in the loop again, because if it is an snapshot, I need to run `as_link` to be able to process it (until we refactor everything to use Snapshot at least). I got rid of the `get_iter` lambda, and shadowed the `all_links` variable (not a fan of shadowing inputs tho): https://github.com/pirate/ArchiveBox/pull/452/commits/4ba3572919b180d1360c4e2f2ffa81be950b5f8a

archivebox/main.py
cdvv7788: https://github.com/pirate/ArchiveBox/pull/452/commits/ee5ba40c831800b83ce2d1e8a82f2cd65422f890

archivebox/main.py
cdvv7788: https://github.com/pirate/ArchiveBox/pull/452/commits/45e6bf555c608325e571104d05e7e778640b2578

archivebox/extractors/__init__.py
pirate: The reason we cannot run `django_setup` at the entrypoint is because some commands can be run before the index is initialized, e.g. `archivebox help`, `archivebox version`, `archivebox init`.

That being said, this function can *only* be called once django is already initialized, so it's safe to depend on django models within it. Where is this function currently being called that doesn't have django already initialized? That callsite should probably be fixed.

cdvv7788: Yes, it is raising the issue. I will check it then.

cdvv7788: @pirate the issue is that `Snapshot` needs to be imported at the module level, and this module is imported at the beginning of `main.py`. Can we bypass the check for now?

cdvv7788:
@pirate I think that most of the requested changes were done. There are a couple of them pending, but we can address them later. What else do you want in place to be able to close this PR?

archivebox/extractors/__init__.py
pirate: Can you use the `if False: import ...` trick to only import it for mypy?

archivebox/index/__init__.py
```diff
@@ -373,7 +372,7 @@ def load_link_details(link: Link, out_dir: Optional[str]=None) -> Link:
     'exact': lambda pattern: Q(url=pattern),
     'substring': lambda pattern: Q(url__icontains=pattern),
     'regex': lambda pattern: Q(url__iregex=pattern),
-    'domain': lambda pattern: Q(domain=pattern),
+    'domain': lambda pattern: Q(url__istartswith=f"http://{pattern}") | Q(url__istartswith=f"https://{pattern}") | Q(url__istartswith=f"ftp://{pattern}"),
```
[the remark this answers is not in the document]
pirate: this workaround seems fine as a semi-long-term solution

pirate (approved):

PR #453, opened 2020-08-20
major refactoring of django views

apkallum:
# Summary

New public index derived from Django db.

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

# Roadmap Goals

Going to 0.5 and deriving all indexes from Django db.


archivebox/core/urls.py
```diff
@@ -31,4 +31,5 @@
     path('index.html', RedirectView.as_view(url='/')),
     path('index.json', static.serve, {'document_root': settings.OUTPUT_DIR, 'path': 'index.json'}),
     path('', MainIndex.as_view(), name='Home'),
+    path('public/', PublicArchiveView.as_view())
```
pirate, line 34: why a separate URL, cant this just be the default homepage when PUBLIC_INDEX=True?

apkallum: Would you like this view to replace the OldHome then?

archivebox/themes/default/core/snapshot_list.html
```diff
@@ -0,0 +1,308 @@
...
+                    iDisplayLength: 100,
+                });
+            });
+        </script>
+        <base href="{% url 'Home' %}">
+    </head>
+    <body data-status="finished">
+        <header>
+            <div class="header-top container-fluid">
+                <div class="row nav">
```
cdvv7788, line 219: This navbar is not present in the `add` view. Given that all of the web interface will focus on django now, we can create actual django templates with a proper structure, having a parent template with the base structure, and overriding the content block with the new snapshot list and add form respectively.

archivebox/core/views.py
```diff
@@ -102,3 +92,61 @@ def get(self, request, path):
...
+    extra_context = {'title': 'Add URLs'}
+
+    def get(self, request, *args, **kwargs):
+        if PUBLIC_ADD_VIEW or self.request.user.is_authenticated:
+            self.extra_context['form'] = AddLinkForm()
+            return render(template_name='add_links.html', request=request, context=self.extra_context)
+        else:
+            return redirect(f'/admin/login/?next={request.path}')
+
+    def post(self, request, *args, **kwargs):
```
cdvv7788, line 128: Missing check for permissions in POST.

Issue #458, opened 2020-08-26
Exception in archive_methods.save_readability


jrruethe:
When attempting to archive [this](https://forums.nasioc.com/forums/showthread.php?t=2803387&amp;highlight=big+brother&amp;page=3) link with Readability enabled, the following exception occurs.

From what I can tell so far, I think the exception occurs somewhere in [here](https://github.com/pirate/ArchiveBox/blob/master/archivebox/extractors/readability.py#L69) before `cmd` is instantiated [here](https://github.com/pirate/ArchiveBox/blob/master/archivebox/extractors/readability.py#L75), which means that `cmd` doesn't exist after the exception is caught and control makes it down to [here](https://github.com/pirate/ArchiveBox/blob/master/archivebox/extractors/readability.py#L109).

```
[√] [2020-08-26 21:10:43] "Big Brother is Watching - '17 WRX Limited and all '17 STI - Page 3 - NASIOC"
    https://forums.nasioc.com/forums/showthread.php?t=2803387&amp;highlight=big+brother&amp;page=3
    √ ./archive/1576899114
      > readability
    ! Failed to archive link: Exception: Exception in archive_methods.save_readability(Link(url=https://forums.nasioc.com/forums/showthread.php?t=2803387&amp;highlight=big+brother&amp;page=3))

Traceback (most recent call last):
  File "/app/archivebox/extractors/__init__.py", line 91, in archive_link
    result = method_function(link=link, out_dir=out_dir)
  File "/app/archivebox/util.py", line 111, in typechecked_function
    return func(*args, **kwargs)
  File "/app/archivebox/extractors/readability.py", line 109, in save_readability
    cmd=cmd,
UnboundLocalError: local variable 'cmd' referenced before assignment

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/archivebox", line 33, in <module>
    sys.exit(load_entry_point('archivebox', 'console_scripts', 'archivebox')())
  File "/app/archivebox/cli/__init__.py", line 122, in main
    run_subcommand(
  File "/app/archivebox/cli/__init__.py", line 62, in run_subcommand
    module.main(args=subcommand_args, stdin=stdin, pwd=pwd)    # type: ignore
  File "/app/archivebox/cli/archivebox_add.py", line 78, in main
    add(
  File "/app/archivebox/util.py", line 111, in typechecked_function
    return func(*args, **kwargs)
  File "/app/archivebox/main.py", line 572, in add
[...]

PR #459, opened 2020-08-27
fix: Add a default cmd

cdvv7788:
# Summary

When the html cannot be retrieved (timeout, unreachable, etc), the cmd value is not set, which causes archivebox to crash. Adding a default value should be enough in that case. 

**Related issues: #458 
# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/extractors/readability.py
+++ b/archivebox/extractors/readability.py
@@ -16,6 +16,7 @@ from ..util import (
 )
 from ..config import (
     TIMEOUT,
+    CURL_BINARY,
     SAVE_READABILITY,
     DEPENDENCIES,
     READABILITY_VERSION,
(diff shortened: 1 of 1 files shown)


archivebox/extractors/readability.py
```diff
@@ -65,6 +66,10 @@ def save_readability(link: Link, out_dir: Optional[str]=None, timeout: int=TIMEO
     # Readability Docs: https://github.com/mozilla/readability
 
     status = 'succeeded'
+    cmd = [
```
pirate, line 69:
```suggestion
    # fake command to show the user so they have something to try debugging if get_html fails
    cmd = [
```

PR #475, opened 2020-09-14
Headers extractor

afreydev:
# Summary

This PR adds a headers extractor.
This get the headers and save them in a headers.json file.

It includes a new get_headers util function, the extractor code and its test.

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/extractors/headers.py
+++ b/archivebox/extractors/headers.py
@@ -1,0 +1,76 @@
+__package__ = 'archivebox.extractors'
+
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+
+from typing import Optional
+import json
+
+from ..index.schema import Link, ArchiveResult, ArchiveError
+from ..system import run, atomic_write
+from ..util import (
+    enforce_types,
+    get_headers,
+    is_static_file,
+)
+from ..config import (
+    TIMEOUT,
+    CURL_BINARY,
+    CURL_USER_AGENT,
+    CURL_VERSION,
+    CHECK_SSL_VALIDITY,
+    DEPENDENCIES,
+)
+from ..logging_util import TimedProgress
+
+@enforce_types
+def should_save_headers(link: Link, out_dir: Optional[str]=None) -> bool:
+    out_dir = out_dir or link.link_dir
+    if is_static_file(link.url):
+        return False
+
+    output = Path(out_dir or link.link_dir) / 'headers.json'
+    return not output.exists()
+
+
+@enforce_types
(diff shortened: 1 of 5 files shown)


archivebox/extractors/headers.py
```diff
@@ -0,0 +1,74 @@
...
+        '--max-time', str(timeout),
+        '--location',
+        '--compressed',
+        '--head',
+        *(['--user-agent', '{}'.format(CURL_USER_AGENT)] if CURL_USER_AGENT else []),
+        *([] if CHECK_SSL_VALIDITY else ['--insecure']),
+        link.url,
+    ]
+    try:
+        json_headers = get_headers(link.url)
```
cdvv7788, line 56: @pirate this is the only detail about this. It is using requests internally to get the headers. Is this ok? (The error `cmd` is falling back to a `curl` command. Or do you prefer to use `curl` for that too?

archivebox/extractors/headers.py
```diff
@@ -0,0 +1,74 @@
...
+    CURL_USER_AGENT,
+    CURL_VERSION,
+    CHECK_SSL_VALIDITY,
+)
+from ..logging_util import TimedProgress
+
+@enforce_types
+def should_save_headers(link: Link, out_dir: Optional[str]=None) -> bool:
+    out_dir = out_dir or link.link_dir
+    if is_static_file(link.url):
```
pirate, line 26: shouldn't it still save the headers even if it's a static file?

archivebox/util.py
```diff
@@ -173,6 +173,20 @@ def download_url(url: str, timeout: int=None) -> str:
 
     return response.text
 
+@enforce_types
+def get_headers(url: str, timeout: int=None) -> str:
+    """Download the contents of a remote url and return the headers"""
+    from .config import TIMEOUT, CHECK_SSL_VALIDITY, WGET_USER_AGENT
+    timeout = timeout or TIMEOUT
+    response = requests.get(
```
pirate, line 181: don't use a get request or it will download the entire response body! use a HEAD request to get only the headers

archivebox/extractors/headers.py
cdvv7788: Yes, I will remove that condition.

archivebox/util.py
cdvv7788: My bad, didn't notice it when I reviewed. @afreydev can you correct it please?

Issue #425, opened 2020-08-07
Question: Tags separation


walkero-gr:
How can I separate multiple tags in an archive. Checked the documentation and didn't find something on that. Maybe I missed it.
A help text on the form would be also useful.

Thanks for your time.

pirate:
Tags are stored as a single comma-separated string for now. They'll be split out into a separate one-to-many foreign-keyed model in a future version.

drpfenderson:
@pirate Is there a way to edit the stored tags on a link in the current version? Or is this more of a future functionality as you mention?

cdvv7788:
@drpfenderson I literally just added that in the PR for the upcoming release. It will be editable in the admin.

pirate:
Closing this because it's finished in the latest release.

PR #479, opened 2020-09-22
Tags migration

afreydev:
# Summary

This PR migrate the snapshot tags field to a new independent model.

# Related #425 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


archivebox/core/admin.py
```diff
@@ -66,6 +66,12 @@ class SnapshotAdmin(admin.ModelAdmin):
     actions = [delete_snapshots, overwrite_snapshots, update_snapshots, update_titles, verify_snapshots]
     actions_template = 'admin/actions_as_select.html'
 
+    def get_queryset(self, request):
+        return super().get_queryset(request).prefetch_related('tags')
+
+    def tag_list(self, obj):
+        return u", ".join(o.name for o in obj.tags.all())
```
pirate, line 73:
```suggestion
        return ', '.join(obj.tags.values_list('name', flat=True))
```

(In python 3 all strings are `u""` strings by default.)

cdvv7788:
LGTM. I fixed a couple issues that were still present. It seems to work pretty well. @pirate can you review it please?

cdvv7788 (approved):

archivebox/core/models.py
```diff
@@ -50,6 +60,13 @@ def as_link(self) -> Link:
     def as_link_with_details(self) -> Link:
         from ..index import load_link_details
         return load_link_details(self.as_link())
+    
+    def get_tags_str(self) -> str:
+        tags = ','.join(
+            tag.name
+            for tag in self.tags.all()
+        ) if self.tags.all() else ''
```
pirate, line 68:
```suggestion
        tags = ','.join(self.tags.values_list('name', flat=True)) if self.tags.exists() else ''
```
minor change, this is a migration so you can ignore it, but for future code this pattern is faster ^

pirate (approved):
Why was the decision made to use `django-taggit` instead of making our own `Tag` model?

I have a mild preference for making our own over using a library, but I assume there was some good reasoning for going the library route.

cdvv7788:
Well, we can definitely change it. A ForeignKey will be better than a GenericRelationship. The base is there, the refactor should not be complex. If you prefer it that way, we can just change it. Given that tags won't be used in any other model, it makes sense.
Taggit has several defaults that make it faster to implement, but other than that there is not a strong reason to keep it around.

Should we change it then?

pirate:
I think I'd like to change it then, as I much prefer the `ForeignKey` to `GenericRelationship`. Sorry for the trouble.

cdvv7788:
No worries. The base is already there, so the refactor should not be complicated.

cdvv7788:
@pirate this requires a m2m relationship, not a ForeignKey. I will still refactor it, but wanted to correct that point.

cdvv7788:
@pirate Please review again. There are some rough edges, but it works well enough.

archivebox/core/models.py
```diff
@@ -50,6 +92,13 @@ def as_link(self) -> Link:
...
```
pirate, line 100: return ','.join(self.tags.values_list('name', flat=True))

PR #481, opened 2020-09-23
Header extractor Improved

afreydev:
# Summary

This PR improve some aspects related to header extractor:

* Replace get method by head.
* Remove check for static files.
# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/config/__init__.py
+++ b/archivebox/config/__init__.py
@@ -85,6 +85,7 @@ CONFIG_DEFAULTS: Dict[str, ConfigDefaultDict] = {
         'SAVE_PDF':                 {'type': bool,  'default': True, 'aliases': ('FETCH_PDF',)},
         'SAVE_SCREENSHOT':          {'type': bool,  'default': True, 'aliases': ('FETCH_SCREENSHOT',)},
         'SAVE_DOM':                 {'type': bool,  'default': True, 'aliases': ('FETCH_DOM',)},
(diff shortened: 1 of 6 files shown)


pirate:
there are a few other things that still need to be addressed before I want to release this extractor in the next version:

- how are 3XX responses / redirects handled? what happens if there are 3 redirects each with their own headers before the final response?
- how are 4XX and 5XX response codes to the HEAD request handled? does it retry with GET?
- what if the response body of the GET request is huge? can we abort the download the moment we get the headers?
- what happens if some of the headers are set-cookie headers? should we use those for the next request in the redirect chain? does `requests.get` handle that?

cdvv7788:
- The headers saved are the last in the chain. @afreydev checked that (there is a missing parameter in the HEAD request, but we are aware and that will get fixed).
- 4xx and 5xx can trigger a retry with GET, that is easy. If the GET still raises that, the returned headers will be saved.
- @afreydev found this https://stackoverflow.com/a/44299915
- https://github.com/psf/requests/issues/1069 According to this issue, requests should handle that.

@pirate It looks like all of the issues are not complex to solve. We will have these changes ready soon.

cdvv7788:
@pirate please check again. I added the required changes, added a `SAVE_HEADERS` flag, and additional tests for this behaviour.

pirate (approved):

PR #489, opened 2020-09-25
migrate config to django settings

apkallum:
# Summary

Move the config to settings. This is still draft as we see where best to do `django_setup` and migrate the settings where they are being imported.  

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [x] Internal architecture

# Roadmap Goals

0.6 roadmap

### DIFF
--- a/archivebox/core/settings.py
+++ b/archivebox/core/settings.py
@@ -1,24 +1,396 @@
 __package__ = 'archivebox.core'
 
 import os
+import re
 import sys
+import json
+import platform
+import shutil
+import getpass
+from typing import Optional
+from pathlib import Path
+from subprocess import run, PIPE, DEVNULL
+from collections import defaultdict
+
+import django
 from django.utils.crypto import get_random_string
 
+import environ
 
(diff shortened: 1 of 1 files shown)


archivebox/core/settings.py
```diff
@@ -1,24 +1,396 @@
...
+HTML_INDEX_FILENAME = 'index.html'
+ROBOTS_TXT_FILENAME = 'robots.txt'
+FAVICON_FILENAME = 'favicon.ico'
+CONFIG_FILENAME = 'ArchiveBox.conf'
+
+# Some values needed to compute other values, so we load them first 
+IS_TTY = sys.stdout.isatty()
+IN_DOCKER= os.getenv("IN_DOCKER", False)
+
+env = environ.Env(
```
pirate, line 192: hmm I'm not convinced that this is better than the current system, doesn't this break tons of stuff like the ini file?

PR #478, opened 2020-09-22
Feature: Add @postlight/mercury-parser as an alternative to Readability

ttimasdf:
# Summary

This PR add [mercury-parser](https://github.com/postlight/mercury-parser) as a backend engine to extract meaningful content from HTML, as an alternative to Readability.

> Mercury Parser powers the Mercury AMP Converter and Mercury Reader, a Chrome extension that removes ads and distractions, leaving only text and images for a beautiful reading view on any site.

mercury parser is also used by [Newsblur](https://github.com/samuelclay/NewsBlur), my favorite RSS Reader 😄 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture
- [x] Archived data layout on disk

# Motivation

I'm trying to archive this article. https://bbs.pediy.com/thread-229215.htm

Readability could only extract the last code block inside this article.

image

image

Mercury Parser is able to extract the whole content

image

PR #492, opened 2020-09-27
Feature: Pre-download images inside HTML content extracted by mercury parser

ttimasdf:
# Summary

This PR added a feature that makes mercury parser somewhere alike readability, but implemented from a different approach. It inline all images into `data:base64,****` form, make the output fully (or partially?) self-contained.

For this moment, readability extract content from `./singlepage.html` variant of archived content. This could be less accurate in some special cases (e.g. the article I mentioned #478 https://bbs.pediy.com/thread-229215.htm) because the HTML is different from original, and the inlined style/media may become a noise. This parser instead find content from `./output.html` which is exactly the same as browser seen, yielding more accurate results. Personally I think it could be a *supplement* to readability parser, not a replacement.

However, some of the media contained (mostly images that I care about) still link to online website/CDN that may expire later, or blocked by hot link protection. This PR add a wrapper script for HTML post-processing, find and extract images from its original URL, and put back to HTML output with its base64-encoded form. It also added a `<meta charset="UTF-8">` at the beginning of content to prevent garbling of non-ASCII content inside some browsers (at least Edge Chromium cannot display correctly both readability and mercury parser output).

This feature is tested manually against this article. at least work for me 😄  https://worldwonderer.github.io/%E9%80%86%E5%90%91%E6%9F%90%E9%9F%B3%E7%9F%AD%E8%A7%86%E9%A2%91App%E4%B9%8B%E8%AE%BE%E5%A4%87%E6%BF%80%E6%B4%BB/

image

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,7 +96,7 @@ ENV IN_DOCKER=True \
     USE_READABILITY=True \
     READABILITY_BINARY="$NODE_DIR/node_modules/.bin/readability-extractor" \
     USE_MERCURY=True \
-    MERCURY_BINARY="$NODE_DIR/node_modules/.bin/mercury-parser"
+    MERCURY_BINARY="$CODE_DIR/bin/mercury-extractor"
 
 # Print version for nice docker finish summary
 RUN archivebox version
--- a/archivebox/extractors/mercury.py
+++ b/archivebox/extractors/mercury.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Optional
 import json
 
+from .readability import get_html
 from ..index.schema import Link, ArchiveResult, ArchiveError
 from ..system import run, atomic_write
 from ..util import (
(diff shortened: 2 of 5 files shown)


pirate:
I see the appeal of this feature, but unfortunately the proposed implementation doesn't work with the way we package archivebox via pip.

The `bin` folder only exists for developers, the final compressed pip package does not include it.

If you make a separate repo for your `mercury-extractor` JS code, we can include it as an npm dependency, the same way we did for `readability-extractor`.

ttimasdf:
I would need some time to get familiar with JS packaging procedures. I seldom write JS myself ;p But `readability-extractor` could be a good starting point, shouldn't be that hard.

pirate:
👍 it's not too difficult but happy to answer any questions. Otherwise let us know in a couple weeks if you lose interest/want us to pick up this PR where you left off. It seems fairly close to the finish line and we wouldn't mind finishing it if you don't get around to it.

pirate:
Thanks for all your work so far on this. I'm going to close it for now because it's getting a bit stale, but if you're still interested in working on it just comment back here and we can discuss how to continue.

PR #493, opened 2020-09-27
Feature: add og:title metadata as alternative title

ttimasdf:
# Summary

[Open Graph Tags](https://www.redclayinteractive.com/what-are-open-graph-tags/) is a mechanism for a website to provide metadata when an URL is shared through social media. I encountered some rare cases that the webpage's `<title>` tag is set through Javascript dynamically at page load, but an `og:title` metadata is present.

This PR make use of open graph tags as a fallback for grabbing page title (also has the potential to gather other metadatas). It extend the `title` extractor with `HTMLParser`, not breaking any existing behaviour. The `<title>` tag content is still the highest priority.

I did not extend the existing regex expression because the HTML is complicated and contains many special cases that cannot be handled easily by regex. In examples below, two attributes in `<meta>` is in different order and only the second one has `<xxx />` form. I didn't even try to parse it through regex. 💫 

examples:

[link2: Blogspot](https://icatch99.blogspot.com/2019/11/icatch-dvr.html)

```html
<meta content="分析台湾可取icatch DVR固件" property="og:title">
```

[link1: WeChat Platform](https://mp.weixin.qq.com/s/KErkNbtnlpaVZRvJd4J-pQ) , 

```html
<meta property="og:title" content="Cocos2dx-js 逆向分析乘凉" />
```

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/extractors/title.py
+++ b/archivebox/extractors/title.py
@@ -1,6 +1,7 @@
 __package__ = 'archivebox.extractors'
 
 import re
+from html.parser import HTMLParser
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


cdvv7788:
Thanks for the PR!
Taking the og tag was part of the plan, so this is aligned with it. The code looks good to me. Can you maybe add some test in `tests/test_extractors.py`? We have a mock server and some templates in there you can use to fetch a title. You could add a small template that has an `og` tag instead to test that case, and have yet another one with no title tag for the case where it returns `None`.

Other than the missing tests, this looks great. @pirate can you have a look at it?

pirate:
I like it, can it fallback to the regex if the HTML parser bails out?  The regex was nice because it also worked on HTML that wasn't valid or had errors, not sure how HTMLParser will handle something like `<html/><head <title>abc` for example.

pirate (approved):

ttimasdf:
> The regex was nice because it also worked on HTML that wasn't valid or had errors

Oh... in that case failover to regex is essential. my fault. I'll add the code back later.

Saying the tests, is there any documentation for setting up the test environment? I may missed something. If i could spin it up, I may also help with writing testcase.

cdvv7788:
The test dependencies are in the `setup.py`. You should be able to install it with `pip install -e .[dev]`. After you have `pytest` and `bottle`, you just need to run `pytest -s` to run the whole test suite. You can check the github actions file, it is configured and run in there.

pirate:
regex fallback added: f727ece

cdvv7788:
I will add the tests in another PR.

Issue #461, opened 2020-08-31
Slow to add new URLs and a lack of verbose logging output


pigmonkey:
I am running ArchiveBox v0.4.21. The archive method section of my config is this:

```
[ARCHIVE_METHOD_TOGGLES]
SAVE_GIT = false
SAVE_MEDIA = false
SAVE_SINGLEFILE = false
SAVE_READABILITY = False
```

Here is the output of adding one new URL from my latest Pinboard JSON dump:

```
$ archivebox add < ~/library/conf/pinboard.json
[i] [2020-08-31 19:26:48] ArchiveBox v0.4.21: archivebox add
    > /home/pigmonkey/tmp/bookmarks

[+] [2020-08-31 19:28:39] Adding 871 links to index (crawl depth=0)...
    > Saved verbatim input to sources/1598902119-import.txt
    > Parsed 871 URLs from input (Generic JSON)
    > Found 1 new URLs not already in index

[*] [2020-08-31 19:30:23] Writing 880 links to main index...
    √ /home/pigmonkey/tmp/bookmarks/index.sqlite3
    √ /home/pigmonkey/tmp/bookmarks/index.json
    √ /home/pigmonkey/tmp/bookmarks/index.html

[▶] [2020-08-31 19:30:29] Collecting content for 1 Snapshots in archive...

[+] [2020-08-31 19:30:29] "welcome the covid influencer - Culture Study"
    https://annehelen.substack.com/p/welcome-the-covid-influencer
    > ./archive/1598900405.0
      > favicon
      > wget
      > pdf
      > screenshot
      > dom
      > archive_org

[√] [2020-08-31 19:32:40] Update of 1 pages complete (2.18 min)
    - 0 links skipped
    - 1 links updated
    - 0 links had errors

    Hint: To view your archive index, open:
        /home/pigmonkey/tmp/bookmarks/index.html
    Or run the built-in webserver:
        archivebox server

[*] [2020-08-31 19:34:26] Writing 880 links to main index...
    √ /home/pigmonkey/tmp/bookmarks/index.sqlite3
    √ /home/pigmonkey/tmp/bookmarks/index.json
    √ /home/pigmonkey/tmp/bookmarks/index.html
```

In total, my shell tells me this took 7 minutes and 43 minutes to complete, which seems exceptionally long for a single URL. There are a few things to note:

[...]

PR #502, opened 2020-10-08
feat: Remove index.json and index.html generation

cdvv7788:
# Summary

At the end of the `cli` commands (add, remove, etc) the json and html indexes were still being generated. This is a slow process dependent on the number of archived links (size of the archive). This PR removes that generation, as they can be generated on demand using the `archivebox list` command.

**Related issues: #461

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [X] Command line interface
- [ ] Configuration options
- [X] Internal architecture

### DIFF
--- a/archivebox/index/__init__.py
+++ b/archivebox/index/__init__.py
@@ -22,7 +22,6 @@ from ..config import (
     ARCHIVE_DIR_NAME,
     SQL_INDEX_FILENAME,
     JSON_INDEX_FILENAME,
-    HTML_INDEX_FILENAME,
     OUTPUT_DIR,
     TIMEOUT,
     URL_BLACKLIST_PTN,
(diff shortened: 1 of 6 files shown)


pirate:
Hmm this PR seems fine but I'd like to make the upgrade path clearer.

Can we create a folder called `exports/` on `archivebox init` and move any `index.html`/`index.json` in the root dir into that folder?
That should make it clear that users need to manually export somehow in order to update those files.
It will also ensure that any tools relying on those files existing in the old location break, forcing people to update those tools to match the new behavior and export process requirements.

cdvv7788:
@pirate I renamed the indexes if they are present. Anything else missing?

pirate (approved):

kedorlaomer:
I liked the `index.html` (although slow generation is clearly bad). Do you think you could add it as an option? A use case (an actual one) is statically hosting the archive directory. This won't work without extra work (basically gathering all files under `archive/*/index.html` and cobbling together an `index.html`)

cdvv7788:
@kedorlaomer you can still generate it using the `list` command. Something like `archivebox list --html --index > index.html` should work.

Issue #503, opened 2020-10-10
Add ability to configure arbitrary additional CLI args for each dependency


pirate:
I'm imagining a feature like this:

`config.py`:
```python
YOUTUBEDL_ARGS = ['-f', 'bestvideo[filesize<500M][height<=?480]+bestaudio/best']
...
WGET_ARGS = ['--no-warc-compression']
...
```

`extractors/youtubedl.py`:
```python
...
CMD = [
    YOUTUBEDL_BIN,
    ...
    *YOUTUBEDL_ARGS,
]
...
```

This will allow people to self-configure the extractor behavior with much more customizability, without requiring archivebox-level config flags for each separate feature.

cdvv7788:
How does this work with the refactor to django's settings? Can it be done at the same time? should we get this one out first?

pirate:
Don't bother with the Django settings refactor for now, I have an idea of how to do it and I'd like to do it myself later on with a module re-organization. Get this one out first as it's quite simple.

cdvv7788:
@pirate we have the following `cmd` in the `media` extractor:
```
cmd = [
        YOUTUBEDL_BINARY,
        '--write-description',
        '--write-info-json',
        '--write-annotations',
        '--write-thumbnail',
        '--no-call-home',
        '--no-check-certificate',
        '--user-agent',
        '--all-subs',
        '--extract-audio',
        '--keep-video',
        '--ignore-errors',
        '--geo-bypass',
        '--audio-format', 'mp3',
        '--audio-quality', '320K',
        '--embed-thumbnail',
        '--add-metadata',
        *(['--yes-playlist'] if SAVE_PLAYLISTS else []),
        *([] if CHECK_SSL_VALIDITY else ['--no-check-certificate']),
        link.url,
    ]
```
The idea is to remove `SAVE_PLAYLISTS` and `CHECK_SSL_VALIDITY` and pass the new `YOUTUBEDL_ARGS` config, right? The default should contain all of these defaults?

cdvv7788:
@pirate I just pushed a commit (https://github.com/pirate/ArchiveBox/pull/506). Can you please check if that is what you had in mind? I will wait until you check it to move forward with other extractors.

cdvv7788:
@pirate Closing this. Please reopen if there is something else we are missing for this.

PR #506, opened 2020-10-14
feat: Add config for extractor args

cdvv7788:
# Summary
Create config options to consolidate extractor args in a single key

# Related issues

#503 
# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [X] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/config/__init__.py
+++ b/archivebox/config/__init__.py
@@ -130,7 +130,12 @@ CONFIG_DEFAULTS: Dict[str, ConfigDefaultDict] = {
                                                                 '--span-hosts',
                                                                 '--no-parent',
                                                                 '-e', 'robots=off',
-                                                                ]}
+                                                                ]},
+        'CURL_ARGS':                {'type': list,  'default': ['--silent',
+                                                                '--location',
+                                                                '--head',
+                                                                '--compressed'
+                                                               ]}
     },
 
     'DEPENDENCY_CONFIG': {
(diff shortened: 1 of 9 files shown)


archivebox/config/__init__.py
```diff
@@ -349,7 +363,10 @@ def load_config_val(key: str,
             raise ValueError(f'Invalid configuration option {key}={val} (expected an integer)')
         return int(val)
 
-    raise Exception('Config values can only be str, bool, or int')
+    elif type is list:
+        return val.split(" ")
```
pirate, line 367: Some arguments have spaces in them that should not be split. It might be better to do `json.loads(val)` and make the user pass lists/objects as json.

pirate (commented):
looks good, can you add `WGET_ARGS`, `CURL_ARGS`, and `CHROME_ARGS`

cdvv7788:
> looks good, can you do the same for `WGET_ARGS`, `CURL_ARGS`, `CHROME_ARGS`, `GIT_ARGS`, etc.

Yes, for sure. I just wanted you to review this first part before adding everything. It is easier to do it this way. I will start fixing it and adding the other args.

cdvv7788:
@pirate chrome_args is actually a function that depends on several things and is used in different ways in different extractors. It is not as easy to change it as it is for the others. I added new config options.

archivebox/config/stubs.py
```diff
@@ -95,6 +94,11 @@ class ConfigDict(BaseConfig, total=False):
     YOUTUBEDL_BINARY: str
     CHROME_BINARY: Optional[str]
 
+    YOUTUBEDL_ARGS: Optional[str]
+    WGET_ARGS: Optional[str]
+    CURL_ARGS: Optional[str]
+    GIT_ARGS: Optional[str]
```
pirate, line 100:
```suggestion
    YOUTUBEDL_ARGS: List[str]
    WGET_ARGS: List[str]
    CURL_ARGS: List[str]
    GIT_ARGS: List[str]
```

pirate (approved):

PR #507, opened 2020-10-18
Add parser for Wallabag Atom feeds

ehainry:
This Pull Request for Atom Feeds generated by Wallabag.

I have not tested how it impacts Shaarli import tool which also parses an Atom feed.


pirate:
thank you! this is great

Issue #496, opened 2020-10-01
Question: what's the current status of the REST API?


zblesk:
I'd like to add new pages by sending an HTTP request to an endpoint. I saw it mentioned in issues such as #339 and items linked in that thread. 

There seemed to be commit names that mentioned adding a REST API, but I haven't been able to find whether those are already implemented and released. 

Are they? If so, how do I call them? 

I've tried just capturing a request to the "Add" method when I click it in the browser, but it looks like there is some csrf protection, so I can't just copy-paste some bearer token and re-issue requests. I'm asking here before I spend time reverse-engineering something just because I missed an already existing API. :)

pirate:
The current status of the API is "unstable" I'd say. Reverse engineering the UI is the way to go for now, but we have plans to stabilize it more in future versions and split out a proper API with django-rest-framework or something so that external tools don't have to shoehorn their needs into requests used by the UI.

mAAdhaTTah:
@pirate I would be interested in working on this. I shot you an email a week or so ago cuz I think the underlying data model needs to be solidified and would love to help move this along. Let me know how I can help.

cdvv7788:
@mAAdhaTTah that is great! Currently, on master, we have the sqlite database working. We can now start working with `django-rest-framework` to enable a proper API (Like @pirate mentioned).
What are the issues that you are finding with the data model? Something that needs to be improved? We can start the discussion here, so we can all have the proper context, and find a way to get started soon.

mAAdhaTTah:
@cdvv7788 Generally, I think the split/transformation between Link <-> Snapshot is a bit weird. Snapshot seems to be db-only (it's transformed into Link's as it's fetched out of the db for most of the operations I was looking at). I also think the double duty of `timestamp` being "the time it was bookmarked" as well as "the path in the archive" is a bit of an issue. From my email:

> I believe you're currently looking to move from timestamp -> sha for the Snapshots and their relationship to the on-disk archive. If we want to eventually allow multiple snapshots per link (to avoid the hash hack), reifiying the Link model into the database and making the Snapshot a single download of a Link seems like a good way to do it. Part of the benefit for me for moving away from timestamps is I want to track when an article was read so I can group them by read day, and manipulating the timestamp for this seems a bit fragile if it can break the relationship to the archive. Having `added`, `updated`, etc. properties for that purpose seems a lot clearer.

[...]

PR #510, opened 2020-10-22
setup_django changes

cdvv7788:
# Summary

Refactor the way django is setup in the application

# Related issues #496 

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [X] Internal architecture

### DIFF
--- a/archivebox/main.py
+++ b/archivebox/main.py
@@ -129,6 +129,14 @@ ALLOWED_IN_OUTPUT_DIR = {
     FAVICON_FILENAME,
 }
 
+
+def using_django(func):
+    def inner(*args, **kwargs):
+        setup_django(out_dir=kwargs["out_dir"], check_db=True)
+        func(*args, **kwargs)
+    return inner
+        
+
 @enforce_types
 def help(out_dir: Path=OUTPUT_DIR) -> None:
     """Print the ArchiveBox help message and usage"""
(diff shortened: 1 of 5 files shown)


archivebox/main.py
```diff
@@ -129,6 +129,14 @@
     FAVICON_FILENAME,
 }
 
+
+def using_django(func):
```
pirate, line 133: this is cool but I think it's too "magic" for not much gain. easier to just keep it as a function call at the top of all the ones that need it.

```python3
def add(urls: Union[str, List[str]], out_dir: Path):
    setup_django(out_dir=our_dir, check_db=True)
    ...
```

pirate:
I have an idea, what if we initialize Django always, but if it's not in a data dir, we use a fake database in memory instead. This also lets us do `oneshot` easily, as we will write to the pretend database and immediately forget it on the next run.

`settings.py`:
```python3
DATABASE_FILE = Path(OUTPUT_DIR) / SQL_INDEX_FILENAME
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': str(DATABASE_FILE) if DATABASE_FILE.exists() else ':memory:',
    }
}
```

cdvv7788:
Superseded by https://github.com/pirate/ArchiveBox/pull/515

PR #518, opened 2020-10-30
tests: Add tests for several different ways to extract the title

cdvv7788:
# Summary

Added the tests for three cases of title extraction:

- `<title>` tag
- og metatag
- `<title>` tag in a malformed html

# Related issues #493 

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


cdvv7788:
@pirate added a couple tests. Also, the parser was not raising an exception. Forcefully raise it now to try to use the HTML regex method in any case if the strict parser fails.

Issue #513, opened 2020-10-24
Move ArchiveResults into the SQL db to avoid checking the filesystem for every Snapshot output status


pirate:
Right now archivebox.zervice.io is taking 6s+ to load because it has to check the filesystem many times for each row on the page during the template rendering process.

We can load the `ArchiveResult`s entries from the `archive/<timestamp>/index.json` files into the sqlite3 db. This will eliminate checking each output path on the filesystem every time we want to see whether it completed or not.

![Screen Shot 2020-10-24 at 3 17 45 PM](https://user-images.githubusercontent.com/511499/97091787-6a134b00-160c-11eb-9d8c-c2beb329c468.png)

cdvv7788:
I will create a model and associate its instances with the `Snapshot` via foreign keys. Is that what you have in mind? @pirate

cdvv7788:
@pirate can this be closed?

PR #520, opened 2020-10-30
Split Snapshot into Link & Snapshot + migrate

mAAdhaTTah:
Proof of concept.

# Summary

Poking around some ideas to improve the data model in ArchiveBox. This introduces splits the Snapshot into a Snapshot & Link model, with a Link representing a single URL added to ArchiveBox, and a Snapshot representing one download that Link at a given time.

There's definitely more work to be done with this (including testing against more than just one URL + tags), but this seems like one of several necessary steps to move data that's currently on the fs into the db (similar to #513).

How does this approach look? Thoughts?

# Related issues

Conversation in #496.

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
Link is actually the last remaining part of the old ArchiveBox core design that we've been moving away from for several months. It's frozen code that only exists as an adapter layer right now, and it's due to be removed completely in the next version in favor of just using Snapshots without a unique URL requirement.

The final portion of that refactor will be timed to land with (or soon after) the move of `ArchiveResult` into the sqlite db, and that will be the last major legacy portion of the codebase to move to the new `Snapshot` system.

This was a reasonable PR for the idea you had, but unfortunately it conflicts with our plans for incrementally changing the core design over the next few months, so I'm going to close it for now. Happy to provide more context and discuss further if we chat this upcoming week (just pick any time on https://calendly.com/nicksweeting/choose-a-time).

mAAdhaTTah:
@pirate Yeah, I went away for the weekend so I haven't had a chance to pick a time. I had a day Friday to do some open source work so thought it would be worth sharing what I was considering in that thread. Will get something on the calendar to figure out where you're going with things and how I can help.

PR #525, opened 2020-11-03
Move ArchiveResult from detail index.json history to database model

cdvv7788:
# Summary

When this PR is ready, archivebox will be able to:
- Save ArchiveResults to the database
- Load ArchiveResults from the filesystem if nothing is found in the database (to ease migration)
- Use ArchiveResults to answer questions about available extractors output per snapshot

# Related issues
https://github.com/pirate/ArchiveBox/issues/513

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [X] Internal architecture


archivebox/core/models.py
```diff
@@ -148,3 +148,13 @@ def save_tags(self, tags=[]):
...
+
+
+class ArchiveResult(models.Model):
+    snapshot = models.ForeignKey(Snapshot, on_delete=models.CASCADE)
+    cmd = models.CharField(max_length=500, default="")
+    pwd = models.CharField(max_length=200, default="")
+    cmd_version = models.CharField(max_length=20, default="")
+    output = models.CharField(max_length=500, default="")
+    start_ts = models.DateTimeField()
+    end_ts = models.DateTimeField()
```
pirate, line 160: also put `extractor = models.CharField(choices=['wget', 'screenshot', ...], ...)`

cdvv7788:
Current state: functional.
It creates `ArchiveResult` objects based on the `index.json` detail for every archive.
It saves new `ArchiveResult` objects when an extractor finish.
The icons displayed in index are calculated based on the `ArchiveResults` for the `Snapshot`

Missing:
 - Remove ArchiveResult schema from the extractor process. Use the ArchiveResult model instead.
 - Add WARC to the list of icons. This requires specific logic.
 - Remove N+1 bottlenecks.
 - Handle the case where there are several `ArchiveResult` objects for a `Snapshot` with the same extractor.

archivebox/core/utils.py
```diff
@@ -1,39 +1,71 @@
...
+        "archive_org": "🏛",
+        "readability": "🆁",
+        "mercury": "🅼",
+    }
+    exclude = ["favicon"]
+    # Missing specific entry for WARC
+
+
+    for extractor in EXTRACTORS:
+        result = archive_results.filter(extractor=extractor[0])
```
pirate, line 29: you should probably also filter out anything that isn't `status='succeeded'` (e.g. status=failed or status=skipped)

archivebox/core/migrations/0007_archiveresult.py
```diff
@@ -0,0 +1,65 @@
...
+        for extractor in history:
+            for result in history[extractor]:
+                ArchiveResult.objects.create(extractor=extractor, snapshot=snapshot, cmd=json.dumps(result["cmd"]), cmd_version=result["cmd_version"], 
+                start_ts=result["start_ts"], end_ts=result["end_ts"], status=result["status"], pwd=result["pwd"], output=result["output"])
+
+
+
+def reverse_func(apps, schema_editor):
+    ArchiveResult = apps.get_model("core", "ArchiveResult")
+    ArchiveResult.objects.all().delete()
```
pirate, line 39: maybe this is too much, but it would be nice to  make sure all the ArchiveResult objects exist in json files before deleting them from DB, otherwise it's potentially destructive

cdvv7788: I will check, and re-create the index.json in case it is not present...just to play safe

archivebox/core/utils.py
```diff
@@ -1,39 +1,75 @@
...
+    for extractor in EXTRACTORS:
+        result = archive_results.filter(extractor=extractor[0], status="succeeded")
+        try:
+            if extractor[0] not in exclude:
+                output += output_template.format(link.archive_path, canon[f"{extractor[0]}_path"],
+                                                 result.exists(), extractor[0], icons.get(extractor[0], "?"))
+            if extractor[0] == "wget":
+                extractor = "warc"
+                output += output_template.format(link.archive_path, canon[f"{extractor}_path"],
+                                                 result.exists(), extractor, icons.get(extractor, "?"))
```
pirate, line 37:
```suggestion
    for extractor, _ in EXTRACTORS:
        result = archive_results.filter(extractor=extractor[0], status="succeeded")
        path, exists = link.archive_path, result.exists()
        try:
            if extractor not in exclude:
                output += output_template.format(path, canon[f"{extractor}_path"], exists, extractor, icons.get(extractor, "?"))
                
            if extractor == "wget":
                # warc isn't technically it's own extractor, so we have to add it after wget
                output += output_template.format(path, canon["warc_path"], exists, "warc", icons["warc"])
```

archivebox/core/migrations/0007_archiveresult.py
cdvv7788: Added a naive check to do this. Should be enough. Let me know if I must improve it further.

cdvv7788:
@pirate Ready for review.

archivebox/core/models.py
```diff
@@ -148,3 +151,18 @@ def save_tags(self, tags=[]):
...
```
pirate, line 163: should `start_ts` and `end_ts` allow `null=True, blank=True`? What happens if the command is run but it never finishes?
 or does that never happen?

archivebox/core/migrations/0007_archiveresult.py
```diff
@@ -0,0 +1,91 @@
...
+            with open(out_dir / "index.json", "r") as f:
+                fs_index = json.load(f)
+        except Exception as e:
+            continue
+
+        history = fs_index["history"]
+
+        for extractor in history:
+            for result in history[extractor]:
+                ArchiveResult.objects.create(extractor=extractor, snapshot=snapshot, cmd=json.dumps(result["cmd"]), cmd_version=result["cmd_version"],
```
pirate, line 33: `cmd=json.dumps(result["cmd"])` why json.dumps here instead of `' '.join()`? should we make it an ArrayField instead so it can hold multiple args natively without string splitting/quoting problems?

archivebox/core/models.py
```diff
@@ -148,3 +151,18 @@ def save_tags(self, tags=[]):
             tags_id.append(Tag.objects.get_or_create(name=tag)[0].id)
         self.tags.clear()
         self.tags.add(*tags_id)
+
+
+class ArchiveResult(models.Model):
+    snapshot = models.ForeignKey(Snapshot, on_delete=models.CASCADE)
+    cmd = models.CharField(max_length=500, default="")
+    pwd = models.CharField(max_length=200, default="")
```
pirate, line 159: Why allow emptystring here for `cmd` and `pwd`? Shouldn't these both always be set?

archivebox/core/models.py
```diff
@@ -148,3 +151,18 @@ def save_tags(self, tags=[]):
...
+    cmd_version = models.CharField(max_length=20, default="")
+    output = models.CharField(max_length=500, default="")
```
pirate, line 161: Same here, I don't think this should ever be empty right? It should always have either a string or an error?

archivebox/core/models.py
```diff
@@ -148,3 +151,18 @@ def save_tags(self, tags=[]):
...
+    start_ts = models.DateTimeField()
+    end_ts = models.DateTimeField()
+    status = models.CharField(max_length=10)
```
pirate, line 164: Let's restrict this to the allowed choices `['succeeded', 'failed', 'skipped']`.

archivebox/core/models.py
```diff
@@ -148,3 +151,18 @@ def save_tags(self, tags=[]):
...
+    extractor = models.CharField(choices=EXTRACTORS, blank=False, max_length=20)
```
pirate, line 165:
```suggestion
    extractor = models.CharField(choices=EXTRACTORS, max_length=32)
```
I was taught that there is no point restricting to anything other than a power of 2^n because PostgreSQL, MySQL, and SQLite all bucket columns to the nearest power of two size.

archivebox/extractors/__init__.py
```diff
@@ -65,6 +66,14 @@ def ignore_methods(to_ignore: List[str]):
 def archive_link(link: Link, overwrite: bool=False, methods: Optional[Iterable[str]]=None, out_dir: Optional[Path]=None, skip_index: bool=False) -> Link:
     """download the DOM, PDF, and a screenshot into a folder named after the link's timestamp"""
 
+    # TODO: Remove when the input is changed to be a snapshot. Suboptimal approach.
```
pirate, line 69: can you clarify this comment? When is the input not a snapshot?

archivebox/themes/default/base.html
```diff
@@ -223,6 +223,10 @@
         .title-col a {
             color: black;
         }
+
+        .exists-False {
+          display: none;
```
pirate, line 228:
```suggestion
          opacity: 0.1;
```

pirate:
This is great so far, sorry I took so long to review it. You don't have to change all the things I asked about in the comments, they're just things I want to understand the reasoning for.

archivebox/core/migrations/0007_archiveresult.py
cdvv7788: I used a json just to keep it as similar to the original ArchiveResult as possible. It is originally a list, and while I could use `join` and `split`, I think that json is more clear on the intent. Also, if there is something like `cmd=["some random", "command"]` the join-split will generate a cmd that is different than the original.

archivebox/core/models.py
cdvv7788: I can remove them. I left it that way in case there were old indexes with those fields empty, but I think I could just ignore those entries if that case arises.

archivebox/core/models.py
cdvv7788: Oh really? Didn't know that! Yes, this can be changed.

cdvv7788:
Thanks for the review! I will fix all of those details today (they are all small details).

archivebox/core/migrations/0007_archiveresult.py
pirate: Why not use `ArrayField` or `JSONField`?

cdvv7788:
@pirate implemented all of the fixes. After you approve it, I will rebase-merge it.

archivebox/core/models.py
```diff
@@ -148,3 +156,18 @@ def save_tags(self, tags=[]):
             tags_id.append(Tag.objects.get_or_create(name=tag)[0].id)
         self.tags.clear()
         self.tags.add(*tags_id)
+
+
+class ArchiveResult(models.Model):
+    snapshot = models.ForeignKey(Snapshot, on_delete=models.CASCADE)
+    cmd = models.CharField(max_length=500)
+    pwd = models.CharField(max_length=200)
```
pirate, line 164:
```suggestion
    pwd = models.CharField(max_length=256)
```

archivebox/core/models.py
```diff
@@ -148,3 +156,18 @@ def save_tags(self, tags=[]):
...
+    cmd_version = models.CharField(max_length=32)
+    output = models.CharField(max_length=500)
```
pirate, line 166:
```suggestion
    output = models.CharField(max_length=512)
```

archivebox/core/models.py
```diff
@@ -148,3 +156,18 @@ def save_tags(self, tags=[]):
...
+    start_ts = models.DateTimeField()
+    end_ts = models.DateTimeField()
+    status = models.CharField(max_length=16, choices=STATUS_CHOICES)
+    extractor = models.CharField(choices=EXTRACTORS, blank=False, max_length=32)
```
pirate, line 170:
```suggestion
    extractor = models.CharField(choices=EXTRACTORS, max_length=32)
```

archivebox/core/utils.py
```diff
@@ -1,39 +1,56 @@
...
+from pathlib import Path
 
 
 def get_icons(snapshot: Snapshot) -> str:
+    archive_results = snapshot.archiveresult_set.filter(status="succeeded")
     link = snapshot.as_link()
     canon = link.canonical_outputs()
-    out_dir = Path(link.link_dir)
+    output = ""
+    output_template = '<a href="/{}/{}" class="exists-True" title="{}">{} </a>'
```
pirate, line 12: I'd like icons to be present for all of the extractors, and greyed out for the ones that are not present. So this should still be `exists-True`/`exists-False` depending on whether or not the successful output is present.

cdvv7788: @pirate please check if something like what I just pushed works for what you have in mind.

archivebox/core/migrations/0007_archiveresult.py
cdvv7788: New in Django 3.1. I will try to bump the django version.

cdvv7788:
@pirate Added two changes you requested. I think this is ready for review.
- Replace charfield with jsonfield
- Re-add icons, even if the related extractors were not used

I also bumped the django version to 3.1

archivebox/core/utils.py
```diff
@@ -1,39 +1,55 @@
...
+
+    for extractor, _ in EXTRACTORS:
+        if extractor not in exclude:
+            exists = extractor_items[extractor] is not None
+            output += output_template.format(path, canon[f"{extractor}_path"], str(exists),
+                                             extractor, icons.get(extractor, "?"))
+        if extractor == "wget":
+            # warc isn't technically it's own extractor, so we have to add it after wget
+            exists = list((Path(path) / canon["warc_path"]).glob("*.warc.gz"))
+            if exists:
```
pirate, line 45: shouldn't this be removed so that the icon is always shown (but with opacity: 0.1 / 1 depending on exists)

pirate (approved):

archivebox/core/utils.py
cdvv7788: yes, missed that. I will push this change directly to master.

archivebox/core/utils.py
```diff
@@ -1,39 +1,55 @@
...
+                output += output_template.format(exists[0], "", str(bool(exists)), "warc", icons.get("warc", "?"))
```
pirate, line 46:
```suggestion
            output += output_template.format(exists[0] if exists else '#', "", str(bool(exists)), "warc", icons.get("warc", "?"))
```

archivebox/core/utils.py
pirate: I fixed it

Issue #233, opened 2019-05-06
PSA: archiving large wallabag collections


anarcat:
## Wiki Page URL

I'm not sure where to put this.. I found a mention of Wallabag in https://github.com/pirate/ArchiveBox/wiki/Usage#CLI-Usage but also https://github.com/pirate/ArchiveBox#can-import-links-from-many-formats

## Suggested Edit

I'm not sure how to phrase this either, but in the former, it says I can export my list of URLs from wallabag using the "Export" button. It doesn't say which format should be used but, in my case, i have over 10 000 links archived in Wallabag, which makes it impractical, to say the least, to export. Any of the buttons just fails with a blank page.

So I made this horrid script to pull all the links into JSON files. It's atrocious, but it works.

```
page=1; while http GET 'https://lib3.net/wallabag/api/entries.json?perPage=100&page='$page 'Authorization:Bearer [REDACTED]' > entries-p$page.json; do
    sleep 1
    page=$(($page +1))
   echo "fetching page $page"
done
```

How to get the Bearer token is explained in "How to create my first app" in the very intuitively named "API clients management" section, e.g. http://wallabag.example.com/developer/howto/first-app

This loop will never complete, because the `http` client is too dumb to return proper exit codes when it hits a 404, and I was too lazy to fix that in script. Just let it run for a while until you start seeing files like this appear:

```
-rw-r--r-- 1 anarcat anarcat        45 2019-05-06 17:23 entries-p180.json
```

They will have this JSON as content:

```
{"error":{"code":404,"message":"Not Found"}}
```

Then the *real* fun begins! While we say archivebox can parse stuff from wallabag, it can't actually parse that JSON!

```
$ archivebox add pages/entries-p2.json 
    > ./sources/entries-p2.json-1557178820.txt

[*] [2019-05-06 21:40:20] Parsing new links from output/sources/entries-p2.json-1557178820.txt...
    > Parsed 0 links as Failed to parse (0 new links added)                                                                                                            

[*] [2019-05-06 21:40:20] Writing 101 links to main index...
    √ /srv/backup/archive/archivebox/index.sqlite3                                                                                                                     
    √ /srv/backup/archive/archivebox/index.json                                                                                                                        
    √ /srv/backup/archive/archivebox/index.html                                                                                                                        

[▶] [2019-05-06 21:40:21] Updating content for 0 matching pages in archive...

[√] [2019-05-06 21:40:21] Update of 0 pages complete (0.00 sec)
[...]

PR #528, opened 2020-11-07
Add parser for Pocket API

mAAdhaTTah:
Pass a url like `pocket://Username` to import that username's archived Pocket
library. Tokens need to be stored in ArchveBox.conf with the following keys:

```
POCKET_CONSUMER_KEY = key-from-custom-pocket-app
POCKET_ACCESS_TOKENS = {"YourUsername": "pocket-token-for-app"}
```

`POCKET_ACCESS_TOKENS` MUST be on a single line, or the JSON will be
misinterpreted by the parser as a new key/value pair.

# Summary

I'm not 100% this is the implementation, but my experience w/ the API is it's more
reliable & complete than the feed export. It would be nice to use this as a feed source
but the last `since` value would need to be persisted somewhere.

# Related issues

None, I wrote this to import my Pocket library locally and was wondering if this would be useful.

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/parsers/pocket_api.py
+++ b/archivebox/parsers/pocket_api.py
@@ -1,0 +1,115 @@
+__package__ = 'archivebox.parsers'
+
+
+import re
+
+from typing import IO, Iterable, Optional
+from datetime import datetime
+from configparser import ConfigParser
+
+from pathlib import Path
+from pocket import Pocket
+import requests
+
+from ..index.schema import Link
+from ..util import (
+    enforce_types,
+)
+from ..config import (
+    SOURCES_DIR
+)
+
+_COUNT_PER_PAGE = 500
+_API_DB_PATH = Path(SOURCES_DIR) / 'pocket_api.db'
+
+# search for broken protocols that sometimes come from the Pocket API
+_BROKEN_PROTOCOL_RE = re.compile('^(http[s]?)(:/(?!/))')
+
+def get_pocket_articles(api: Pocket, since=None, page=0):
+    body, headers = api.get(
+        state='archive',
+        sort='oldest',
+        since=since,
+        count=_COUNT_PER_PAGE,
+        offset=page * _COUNT_PER_PAGE,
+    )
+
+    articles = body['list'].values() if isinstance(body['list'], dict) else body['list']
+    returned_count = len(articles)
+
+    yield from articles
+
+    if returned_count == _COUNT_PER_PAGE:
+        yield from get_pocket_articles(api, since=since, page=page + 1)
+    else:
+        api.last_since = body['since']
+
+
+def link_from_article(article: dict, sources: list):
+    url: str = article['resolved_url'] or article['given_url']
+    broken_protocol = _BROKEN_PROTOCOL_RE.match(url)
+    if broken_protocol:
+        url = url.replace(f'{broken_protocol.group(1)}:/', f'{broken_protocol.group(1)}://')
+    title = article['resolved_title'] or article['given_title'] or url
+
+    return Link(
+        url=url,
+        timestamp=article['time_read'],
+        title=title,
+        tags=article.get('tags'),
+        sources=sources
+    )
+
+def write_since(username: str, since: str):
+    from ..system import atomic_write
+
+    if not _API_DB_PATH.exists():
+        atomic_write(_API_DB_PATH, '')
+
+    since_file = ConfigParser()
+    since_file.optionxform = str
+    since_file.read(_API_DB_PATH)
+
+    since_file[username] = {
+        'since': since
+    }
+
+    with open(_API_DB_PATH, 'w+') as new:
+        since_file.write(new)
+
+def read_since(username: str) -> Optional[str]:
+    from ..system import atomic_write
+
+    if not _API_DB_PATH.exists():
+        atomic_write(_API_DB_PATH, '')
+
+    config_file = ConfigParser()
+    config_file.optionxform = str
+    config_file.read(_API_DB_PATH)
+
+    return config_file.get(username, 'since', fallback=None)
+
+@enforce_types
+def should_parse_as_pocket_api(text: str) -> bool:
+    return text.startswith('pocket://')
+
+@enforce_types
+def parse_pocket_api_export(input_buffer: IO[str], **_kwargs) -> Iterable[Link]:
+    """Parse bookmarks from the Pocket API"""
+
+    input_buffer.seek(0)
+    pattern = re.compile("^pocket:\/\/(\w+)")
+    for line in input_buffer:
+      if should_parse_as_pocket_api(line):
+        from ..config import (
+          POCKET_CONSUMER_KEY,
+          POCKET_ACCESS_TOKENS,
+        )
+        username = pattern.search(line).group(1)
+        api = Pocket(POCKET_CONSUMER_KEY, POCKET_ACCESS_TOKENS[username])
+        api.last_since = None
+
+        for article in get_pocket_articles(api, since=read_since(username)):
+          yield link_from_article(article, sources=[line])
+
+        write_since(username, api.last_since)
--- a/archivebox/config.py
+++ b/archivebox/config.py
@@ -159,6 +159,9 @@ CONFIG_DEFAULTS: Dict[str, ConfigDefaultDict] = {
         'YOUTUBEDL_BINARY':         {'type': str,   'default': 'youtube-dl'},
         'NODE_BINARY':              {'type': str,   'default': 'node'},
         'CHROME_BINARY':            {'type': str,   'default': None},
+
+        'POCKET_CONSUMER_KEY':      {'type': str,   'default': None},
+        'POCKET_ACCESS_TOKENS':     {'type': dict,  'default': {}},
     },
 }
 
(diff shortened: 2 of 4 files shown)


pirate:
This looks great. I think it's reasonable to get the Pocket API credentials from config, let me know if you want help with doing that.

Also side note: I thought about this for a while, and I think I actually really like your `pocket://Username` pattern for explicitly specifying how to parse an input. Using this pattern more would save the hassle of having to try so many different parsers for each input. This way we could even do `history://Default@Chrome` or `bookmarks://Default@Firefox` just like how we allow `http://...` and `ftp://`.

Would be great to have 2 functions for each parser `should_parse_rss(text)` and `parse_rss(text)`, just like we have for the extractors with `should_save_wget()` and `save_wget()`.

```python
def should_parse_as_pocket_api(text):
    return text.startswith('pocket://')

def parse_pocket_api(text):
    ...
```

mAAdhaTTah:
> This looks great. I think it's reasonable to get the Pocket API credentials from config, let me know if you want help with doing that.

Yeah, not 100% sure on this part yet, so any help you could provide would be appreciated.

> Would be great to have 2 functions for each parser should_parse_rss(text) and parse_rss(text), just like we have for the extractors with should_save_wget() and save_wget().

To clarify: Is the idea here to modify all of the extractors to have these two functions and update the loop to use them?

> Also side note: I thought about this for a while, and I think I actually really like your pocket://Username pattern for explicitly specifying how to parse an input.

Setting aside for a second how this would apply to the other extractors, I have to admit the `pocket://Username` setup is a bit of a lie. The extractor (perhaps obviously) does use the `pocket://` protocol but doesn't actually use the Username part. Pocket will provide a Consumer key (I think that's what it was called?) for the Pocket app you create, then the individual user need to provide the matching Access key (?) after logging into the app. I have these two already, so I just filled them into the spots in the constructor, but there's no mechanism for getting an Access key and I'm not 100% sure how we should do that. It's a fairly straightforward OAuth flow, and we could put the Consumer key in the config file, but this begs two questions:

1. How would we implement the OAuth flow in a semi-consistent way for both the CLI & web app?
2. Where should we store the Access tokens after completing the OAuth flow?

I will say, one of the thoughts I had was to pass the Access key as the username, so `pocket://long-uuid-for-access-token`, but it would "leak" the access token into the logs, which probably isn't great.

I will also mention the overall approach here could be useful for other extractors like Wallabag (related: #233), so could be useful to consider this somewhat generically.

Thoughts on all of this?

mAAdhaTTah:
This implementation works but the JSON string w/ `configparser` is a bit meh, but conceptually, this is what I'm proposing for the Pocket API implementation. If anyone has a better idea for putting this data in the config file, I'm open, but this works.

archivebox/parsers/pocket_api.py
```diff
@@ -0,0 +1,120 @@
...
+def should_parse_as_pocket_api(text: str) -> bool:
+    return text.startswith('pocket://')
+
+@enforce_types
+def parse_pocket_api_export(html_file: IO[str], **_kwargs) -> Iterable[Link]:
+    """Parse bookmarks from the Pocket API"""
+
+    html_file.seek(0)
+    pattern = re.compile("^pocket:\/\/(\w+)")
+    for line in html_file:
```
pirate, line 106:
```suggestion
def parse_pocket_api_export(input_buffer: IO[str], **_kwargs) -> Iterable[Link]:
    """Parse bookmarks from the Pocket API"""

    input_buffer.seek(0)
    pattern = re.compile("^pocket:\/\/(\w+)")
    for line in input_buffer:
```

mAAdhaTTah:
@pirate Is this looking good otherwise? Not 100% sure how I feel about the `read_since` & `write_since` setup, so interested in your thoughts there, although it works well enough so far in my experience.

pirate:
Yeah this looks great so far. We have quite a lot of things to merge in this release, so it may take me a while to get around to merging and testing this fully. Thanks for the work so far and your patience though.

mAAdhaTTah:
@pirate Awesome, good to know, thanks. I've got a few other ideas & things I'm going to send your way as well, feel free to review/merge them as you have the time, and I'll pull them together in a branch in my fork until they get upstreamed.

cdvv7788:
The config system looks promising...I will upload something tomorrow to test the `wget` extractor with this new structure.

archivebox/parsers/pocket_api.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if returned_count == _COUNT_PER_PAGE:
+        yield from get_pocket_articles(api, since=since, page=page + 1)
+    else:
+        api.last_since = body['since']
+
+
+def link_from_article(article: dict, sources: list):
+    url: str = article['resolved_url'] or article['given_url']
+    if url.startswith('http:/www'):
```
pirate, line 45: huh, I haven't seen this type of mangled URL from pocket before. if it's a pocket API bug of some kind producing these, should we do the same check for https?

archivebox/parsers/pocket_api.py
```diff
@@ -0,0 +1,119 @@
...
+from pocket import Pocket
+import requests
+
+from ..index.schema import Link
+from ..util import (
+    enforce_types,
+)
+
+_COUNT_PER_PAGE = 500
+_SINCE_FILENAME = 'since.db'
```
pirate, line 20: can you move the config import to the top of the file and set the full path here

```suggestion
_SINCE_FILENAME = Path(SOURCES_DIR) / 'pocket_api.db'
```

archivebox/parsers/pocket_api.py
mAAdhaTTah: True, will add that.

mAAdhaTTah:
@pirate Fixed everything & ready for re-review.

archivebox/parsers/pocket_api.py
```diff
@@ -0,0 +1,115 @@
...
+from ..index.schema import Link
+from ..util import (
+    enforce_types,
+)
+from ..config import (
+    OUTPUT_DIR
+)
+
+_COUNT_PER_PAGE = 500
+_SINCE_FILENAME = Path(OUTPUT_DIR) / 'pocket_api.db'
```
pirate, line 23: I think it makes sense to put it in the SOURCES_DIR subfolder, not the root dir.

pirate (approved):

archivebox/parsers/pocket_api.py
mAAdhaTTah: Done.

pirate:
sorry for the rebase but I had to point this to the v0.5.0 draft branch instead of master, do you mind fixing the conflicts and then I'll merge it

mAAdhaTTah:
@pirate Happy to update it but wondering if we need to wait for the config changes cuz the config file might have to change with the new parser to handle the API key dict.

pirate:
Nah don't worry about it we'll change it when we merge that PR.

mAAdhaTTah:
@pirate ok cool updated.

pirate:
ah shoot I didn't catch earlier that this added a new python library dependency.

Unfortunatley we can only add python libraries that are also apt packages, due to the restrictions on how debian packages are distributed.
This means every python library we add like `pip install pocket` needs to have an equivalent `apt install python3-pocket` package available in the mainline Ubuntu/Debian repos.

The alternative is to vendor the entire pocket library inside ArchiveBox, but we generally try not to do that if it can be avoided.

I may have to pull this out of the v0.5.0 release for now until we can figure out a solution.

mAAdhaTTah:
@pirate Did you delete your comment? If the pocket lib is an issue, I can refactor to use requests directly. The lib isn't so extensive.

pirate:
Yeah because I already fixed it by vendoring the lib. iIt's only a single file so it's fine, I was only worried about vendoring if it was a big complicated Python lib.

filviu:
Sorry to resurect this - but I couldn't find any documentation on setting up Pocket API other than this PR and #726

I obbtained my keys and I can use the API using CURL so I'm fairly confident I have correct key and token.

Running 

```bash
archivebox config config --set POCKET_CONSUMER_KEY=aaaaaa-0123455....
```
was straightforward but no matter of quoting and escaping allowed me to run `--set POCKET_ACCESS_TOKENS` I would either get JSON Parser errors or `[X] Config KEY=VALUE must have an = sign in it`

I tried manually editing ArchiveBox.conf but I don't think I nailed the syntax, running `--get POCKET_ACCESS_TOKENS` returns nothing and running add fails:

```bash
archivebox add --parser pocket_api pocket://MyUserName
                                                                                                                                                           0.4% (1/240sec)
[X] No links found using Pocket API parser
    Hint: Try a different parser or double check the input?
```

My ArchiveBox.conf looks like this:

```bash
bash-3.2# grep POCKET ArchiveBox.conf
POCKET_CONSUMER_KEY = 123456-xxxxxxxxxxxxxxxxx
POCKET_ACCESS_TOKENS = {"myusername":"12345678-aa00-aa00-aa00-aa0000"}
```

mAAdhaTTah:
@filviu Main thing that jumps out to me is `myusername` & `pocket://MyUserName` aren't the same capitalization. Dunno if that's caused by the example or what, but def make sure those are the same.

filviu:
My example was unfortunate, I'm using the proper username in both places (no uppercase at all).

PR #529, opened 2020-11-08
Draft of new REST API using DRF

mAAdhaTTah:
This pulls in DRF to configure our API. Pretty straightforward binding of a view
to a serializer & a model and making the data available. For this first pass,
we're using the model even though it's currently unstable. From a feature
standpoint, we get a lot for free from DRF with very little code, including
pagination. The `list_links` method loads all of the snapshots, which would
require pagination to be implemented manually on the entire list of snapshots,
which won't work well on large databases. Because archivebox is a CLI first
and a web application second, the way Exceptions are thrown and errors
logged doesn't always make those methods conducive to integrating w/ an API.

On the testing side, this shows up in how we're configuring things. The
`setup_django` function doesn't fully work when passing `out_path`;
Some variables in the Django settings aren't updated or configured correctly.
Instead, we use `subprocess` the same way the other tests do to start up the
server and hit it with `requests`.

# Summary

This is obviously a work in progress but wanted to get some feedback on the
direction. It would be helpful if the API functions exposed by archivebox were
more decoupled from the CLI context specifically, but I think we're going to
want to bind the Models directly (at least for querying).

# Related issues

#496 

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox.egg-info/SOURCES.txt
+++ b/archivebox.egg-info/SOURCES.txt
@@ -21,6 +21,9 @@ archivebox.egg-info/dependency_links.txt
 archivebox.egg-info/entry_points.txt
 archivebox.egg-info/requires.txt
 archivebox.egg-info/top_level.txt
+archivebox/api/__init__.py
+archivebox/api/serializers.py
+archivebox/api/views.py
 archivebox/cli/__init__.py
 archivebox/cli/archivebox_add.py
 archivebox/cli/archivebox_config.py
--- a/archivebox.egg-info/requires.txt
+++ b/archivebox.egg-info/requires.txt
@@ -4,6 +4,7 @@ mypy-extensions==0.4.3
 base32-crockford==0.3.0
 django==3.0.8
 django-extensions==3.0.3
+djangorestframework==3.12.2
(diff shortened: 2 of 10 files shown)


pirate:
This looks great! It's going in a good direction and the implementation looks concise and straightforward so far. Thanks for doing the test setup too.

mAAdhaTTah:
(Didn't mean to make ready.)

What's the semantic meaning of the `added`, `updated`, & `bookmarked` (which is derived from `timestamp`, which I believe is kind of like `added` but for Pocket, both API + export, it's the time the link was archived)? Is there a goal of removing one of these? Should these be customizable via the API? e.g. it kinda looks like the `timestamp`, and thus `bookmarked` can be set via the Pocket import; should the REST API enable that too?

pirate:
Going to close this as stale. Thanks for all your work on this so far.

I'm planning to use https://django-ninja.rest-framework.com/ going   forward for the REST API (instead of DRF).

PR #537, opened 2020-11-14
File icons improvement

TrAyZeN:
# Summary

This PR fixes the order of file icons and changes the opacity of file icons on the public view.

# Related issues

# Changes these areas

- [x] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


cdvv7788:
@TrAyZeN you will need to change your PR to be based on https://github.com/pirate/ArchiveBox/pull/525 . That PR changes a lot of internal stuff, including the way the icons are included.

pirate:
Why reorder them? I prefer the old order because it's ordered from most useful -> least useful, and the DOM dump is one of the worst snapshot formats in terms of reliability.

TrAyZeN:
Sorry for answering that late, I didn't know there was a prefered order but this PR was not about the order. The problem was that the icons were not corresponding to the links. For instance, when you clicked on the 💻 icon it would link to html dump instead of screenshot.

pirate:
Ah, thanks for clarifying, I didn't see that earlier :+1:

PR #542, opened 2020-11-19
Settings refactor using dynaconf

cdvv7788:
# Summary

Replace the current config structure with a tiered architecture to ease readability and composability of the configurations around the application.
In this case there are several options:
- Load the main configuration and override values that are found in the local configuration
- Load the main configuration and use it as the basis to calculate values in the local configuration

I went with the second option, because it is less traumatic at this point. There are a lot of values that are calculated at runtime, so it is not as easy as dropping some config files and get done with it. If we are still going to require some logic to calculate some `args`, at least now we can keep them contained in the config files (mostly). For example, to calculate the version of some binaries, that is done at the upper level. We cannot just inherit it as a plain setting. We could move all of the configurations that require these calculations down to the extractors, but that would make this refactor way bigger.

So far, the `wget` extractor seems to work with this new structure. It is now on its own module with its own settings, and the logic dedicated to generate the flags to pass to the command is in the configuration script now. The actual extractor looks simpler now. There are some hardcoded values that will need to be replaced if we decide we want to keep going this way.

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [X] Configuration options
- [X] Internal architecture


mAAdhaTTah:
The secrets implementation + the nested config would be really helpful for #528. Exciting!

pirate:
Going to close this PR since it's very stale, moving discussion to an issue: https://github.com/ArchiveBox/ArchiveBox/issues/693

Issue #22, opened 2017-06-11
Add ability to do full-text search of archived text/markdown/html content


ilvar:
As a user with lots of bookmarks, I want to have a search input to filter list by text matches in URL, title and extracted content. 

http://elasticlunr.com/ can be used for simple client side full text search.

pirate:
This is definitely one of my goals, but think [`ag`](https://github.com/ggreer/the_silver_searcher) is more than fast enough for 90% of people, and it supports regex.
I just whipped this search script up, all that needs to be done now is add a search field to the index.html which queries the server and highlights matching sites.

**Usage:**

1. set `ARCHIVE_PATH` to your actual archive path
2. use it on the cli
```bash
./search.py "train number \d+"     # search with regex
./search.py --exact "[1] (citation needed)."   # search for an exact string
```
3. use it as a REST service
```bash
./search.py --server 8080
```
`http://127.0.0.1:8080/search?search=firefox` ->
```
/1409779227/www.mozilla.org/en-US/firefox/help/index.html
/1409779227.1/www.mozilla.org/en-US/firefox/customize/index.html
/1409779227.3/www.mozilla.org/media/img/firefox/horizon/stars.8398dac91f60.svg
/1497562974/www.ghacks.net/2009/07/23/firefox-bookmarks-archiver/index.html
/1497562974.2/duckduckgo.com/index.html?q=firefox+export+bookmarks&t=ffhp&ia=web.html
```

**search.py:**
```python
import sys
from subprocess import run, PIPE

ARCHIVE_PATH = '/Users/yourusername/Documents/bookmark-archiver/bookmarks/archive'

if run(['which', 'ag'], stderr=DEVNULL).returncode:
    print("[X] Please install ag the silver searcher:\n\t apt install silversearcher-ag\n\t brew install the_silver_searcher")
    raise SystemExit(1)

def search_archive(pattern, regex=False):
    args = '-g' if regex else '-Qg'
    ag = run(['ag', args, pattern, ARCHIVE_PATH], stdout=PIPE, stderr=PIPE, timeout=60)
    ls = run(['ls', ARCHIVE_PATH], stdout=PIPE, stderr=PIPE, timeout=60)
    return (l.decode().replace(ARCHIVE_PATH, '') for l in ag.stdout.splitlines())


def server(port=8080):
    try:
        from flask import Flask
        from flask import request
    except ImportError:
        print('[X] Please install Flask to use the search server: pip install Flask')
[...]

PR #543, opened 2020-11-19
Full-text search

jdcaballerov:
# Summary

This PR Adds the ability to do full-text search 🎉 

# Related issues
 #22 #24 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [x] Internal architecture


jdcaballerov:
Will change to use generators for the ids.

cdvv7788:
@jdcaballerov LGTM! Can you please update the `Dockerfile` so it is installed automatically? This only provides FTS in the admin, right? Can we add something in the `list` command too?

jdcaballerov:
@cdvv7788 will sonic be a required dependency ? Right now only admin, I'll check tomorrow how we could integrate it with `list`.

cdvv7788:
Not necessarily. You can leave it as the first option, and use the current mechanism as a fallback. I just want to have it in the cli where we can test that it actually works without digging into the code.

jdcaballerov:
@cdvv7788 Could you please be more specific about the needs for `list`. The filtering currently in place in the django admin is not comparable to `list` filtering.

1. `list` command chains the following filters
```python
LINK_FILTERS = {
    'exact': lambda pattern: Q(url=pattern),
    'substring': lambda pattern: Q(url__icontains=pattern),
    'regex': lambda pattern: Q(url__iregex=pattern),
    'domain': lambda pattern: Q(url__istartswith=f"http://{pattern}") | Q(url__istartswith=f"https://{pattern}") | Q(url__istartswith=f"ftp://{pattern}"),
}```
while the django admin uses this search fields
2.
```python
    search_fields = ['url', 'timestamp', 'title', 'tags__name']
```

jdcaballerov:
Behavior proposal:

## ADMIN
Search enabled

Screenshot from 2020-11-19 23-42-12


Search enabled but backend fail

Screenshot from 2020-11-19 23-43-28

## CLI

New search filter backend enabled all OK

Screenshot from 2020-11-19 23-45-14

New search filter backend enabled but fail

Screenshot from 2020-11-19 23-46-45


New search filter backend disabled (fail and ok)

Screenshot from 2020-11-19 23-52-11

cdvv7788:
Yes, I had in mind a new filter. Where is it searching? Using readability?

jdcaballerov:
@cdvv7788 currently I've only added readability content but any texts from the extractors could be added. i.e each one of the headers, etc.

What is needed is to add a `list` in `index_texts` at the extractor returned `ArchiveResult` dataclass. If `index_texts` is not present it will only pass.

```python

    return ArchiveResult(
        cmd=cmd,
        pwd=str(out_dir),
        cmd_version=READABILITY_VERSION,
        output=output,
        status=status,
        index_texts= [readability_content] if readability_content else [],
        **timer.stats,  
    )
```

jdcaballerov:
@cdvv7788 
Sonic is a service with interaction via a telnet like protocol, can't be added to the `Dockerfile` unless one includes supervisord to manage the container.

I've added sonic to the docker-compose file as a service. Please don't forget to uncomment the line `build .` and comment the next line so that it doesn't use the image from dockerhub.

pirate:
How do you handle retroactively adding previous archive data to the index? When a user upgrades to this release will it start indexing all their previously archived text?

jdcaballerov:
@pirate Retroactively is not doing anything. The indexing is taking place in `archivebox/extractors/__init__.py.archive_link`
if indexing is enabled (`config.USE_INDEXING_BACKEND`) 

I thought someone wanting indexing will run `update`

pirate:
Hmmm we definitely need some solution that can index >50k articles with ~2000 words each, and it needs to start indexing retroactively somehow when its enabled. Archivebox update will not re-run extractors that have already run, so I don't think that's enough to index everything.

Are there any sonic alternatives you know of that can handle indexing 100m - 2bn words?

ekiel:
Is there a way to integrate recoll? https://www.lesbonscomptes.com/recoll/

Eric Kiel

pirate:
Recoll seems pretty heavyweight, ideally I'm looking for something similar to `sonic`, just with a bigger index ability.

Side note, it would be great to offer `ripgrep` as a search backend fallback @jdcaballerov (e.g. in Docker) if sonic or another index-searching backend is not available. Ripgrep is faster and simpler for smaller archives and can be installed as a static binary and called when needed, instead of needing a constantly running backend.

jdcaballerov:
@pirate `sonic` might be able to handle the requirements

query_limit_maximum can go up to (u16) - 65535
retain_word_objects (usize) can go up to 18_446_744_073_709_551_615 on some systems

but I think the question is about being exhaustive or not, if it's ok with max 65535 results per query and a word related to maximum X articles we are ok.

The retroactive indexing can be figured out.

jdcaballerov:
I've added ripgrep (rg) backend and set it as default.

jdcaballerov:
Added the ability to do retroactive indexing using `archivebox update --index-only`. Will select the content from the first available extractor output registered as a `succeeded` `ArchiveResult` in the order (readability, singlefile, dom, wget) and execute indexing in the current search backend (ripgrep doesn't do anything).

ps;
--index-only still needs to be corrected

pirate:
bump for rebase, would love to get this out with v0.5.0

pirate:
Fixed the conflicts and merged this here: https://github.com/ArchiveBox/ArchiveBox/pull/570

PR #548, opened 2020-11-23
Parse Snapshot#timestamp with `fromtimestamp`

mAAdhaTTah:
Since we know what this is supposed to be, we don't need to use the "best
guess" version from `parse_date`. Some values I was getting from the Pocket
API produced weird/inconsistent results.

# Summary

I got some weird results with some timestamps from the Pocket API, which
_could_ just be issues in Pocket's data, but as I was trying to solve them, I
realized we had this "best guess" approach for parsing a date, while we
know for sure this is a timestamp and could be parsed as one.

# Related issues

N/A but wondering if maybe `bookmarked` should be reified into the DB?
Would be helpful for the REST API, so we could sort by bookmarked.

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/core/models.py
+++ b/archivebox/core/models.py
@@ -1,6 +1,7 @@
 __package__ = 'archivebox.core'
 
 import uuid
+from datetime import datetime
 
 from django.db import models, transaction
 from django.utils.functional import cached_property
@@ -98,7 +99,7 @@ class Snapshot(models.Model):
 
     @cached_property
     def bookmarked(self):
-        return parse_date(self.timestamp)
+        return datetime.fromtimestamp(float(self.timestamp))
 
     @cached_property
     def is_archived(self):


pirate:
Why do we know for sure that self.timestamp is a valid timestamp? It's not normalized on import and often it's taken as whatever the import source provides verbatim.

This is why we maintain `self.added` and `self.updated` separately which are guaranteed to be valid timestamps (with tzinfo) produced and consumed by django. `bookmarked` is the name for the attempted parsed version of `self.timestamp`, but it doesn't always succeed and the timestamp parsing logic changes over time as we improve it, so it's important to always keep the original string version and any attempted parsed version separate.

mAAdhaTTah:
> It's not normalized on import

After looking through the parsers, it is in all of them. They mostly do `datetime.fromtimestamp` before setting it to `Link#timestamp` (with other cases like `generic_json` doing a whole set of attempts to get a `timestamp`). But ultimately, it ends up being a `timestamp` (which is probably something I should fix in the the Pocket API PR).

> it's important to always keep the original string version and any attempted parsed version separate.

FWIW, I'd be in favor of splitting these into two different columns in the db. `timestamp` is doing double-duty as both the time when the item was added/bookmarked/whatever-the-source-wants as well as the folder in `data/archive`. If we moved this into an actual `bookmarked` date field in the db, `timestamp` would just be the one thing.

Then maybe we could move `data/archive` to use `sha` and kill `timestamp` completely, and make it much clearer what those properties actually mean. I had mentioned this in the REST API thread, because then we could more easily query `bookmarked` by date rather than than timestamp without needing to do a bunch of translation either in the API or (yuck) the client.

pirate:
> After looking through the parsers, there is normalization in all of them

This is a relatively new addition, in the past we did not normalize these and there are users with thousands of links with invalid timestamps. For backward compatibility reasons we can never assume timestamps are always valid datetimes.

> Then maybe we could move data/archive to use sha and kill timestamp completely

This is already planned and in-progress. We're using a truncated base32 hash of the URL for the v0.7 migration from timestamps to sha folders.
Backwards compatibility is being handled by the Django server accepting either the timestamp or base32 hash to retrieve a given snapshot, so only the on-disk layout will break.

mAAdhaTTah:
@pirate Ah well. Good to know the history.

What's your feeling on adding `bookmarked` (or maybe a different field name like `saved`?) to the db as a datetime that the API could use?

pirate:
I'm still hesitant to put `bookmarked` in the database, as so many of the timestamps can be incorrect and misleading for some users. I fear that if we add it as a queryable field people will start writing code that uses it, and those features will be built on a foundation of bad data. It's especially insidious because in pristine dev environments with test data everything looks fine, the problem is only apparent when working with large historical archives that have been imported over years or from particularly misbehaved browsers. It's easy for contributors to think it's a stable datetime source and not realize that it has a complicated history.

I'd rather we usher people towards using only `added`/`updated` for DB queries. I could maybe still be convinced to add it depending on the reason, is there a particular feature you had in mind that you wanted to use `bookmarked` for?

mAAdhaTTah:
> I'd rather we usher people towards using only added/updated for DB queries, but I could maybe be convinced to add it depending on the reason, is there a particular feature you had in mind that you wanted to use bookmarked for?

Yeah, the import from Pocket uses the `archived` timestamp from the Pocket API to populate the `timestamp` in ArchiveBox, which conveniently maps to the time I finished reading the piece, which is how I want to organize my link display. Using added / updated makes it look like I read 27k articles within the same 10s in 2020 :D. So I need something that maps from the archive source into ArchiveBox, which I can then query off of in the REST API for my site.

mAAdhaTTah:
Open question here about reifying `bookmarked` into the db but obvs this PR is not gonna work.

PR #552, opened 2020-11-28
v0.5.0 Release

pirate:
- ArchiveResult moved to SQLite3 DB for performance @cdvv7788 
- lots of assorted bugfixes and improvements courtesy of @cdvv7788 and @jdcaballerov
- new full-text search support with ripgrep and sonic courtesy of @jdcaballerov
- new `archivebox oneshot` command for downloading a single site without starting a whole collection
- new Pocket API importer courtesy of @mAAdhaTTah
- new Wallabag importer courtesy of @ehainry
- new extractor options on Add page courtesy of @BlipRanger
- new apt/deb/homebrew/pip packaging setup into separate repos under new Github Org https://github.com/ArchiveBox
- new official PPA and Docker Hub accounts https://hub.docker.com/r/archivebox/archivebox (with automatic armv7 builds courtesy of @chrismeller)


jdcaballerov:
@pirate The only thing I noticed is that the sonic client dependency is missing. Sonic backend is not the default however.

https://github.com/jdcaballerov/ArchiveBox/blob/796d5d7fc12008bdfd7ceaa623da2aea6c23b3e9/archivebox.egg-info/requires.txt#L11

jdcaballerov (approved):
@pirate The only thing I noticed is that the sonic client dependency is missing. Sonic backend is not the default however.

Original commit:
https://github.com/jdcaballerov/ArchiveBox/blob/796d5d7fc12008bdfd7ceaa623da2aea6c23b3e9/archivebox.egg-info/requires.txt#L11

pirate:
Every python library added needs to have an apt-installable equivalent, or be vendored. It's a requirement for debian packaging, we can only depend on other debian packages, not pip packages, so every pip package also needs to be a debian package.

If you an `apt install python3-sonic-client` then we can add it in `setup.py`, `stdeb.cf`, and `homebrew-archivebox/archivebox.rb`.

If not we need to vendor the entire sonic-client library inside of `archivebox/vendor/`, or make it optional and not install by default.

PR #563, opened 2020-12-02
fix "inconsisntencies" typo in error message

hawkrives:
# Summary

I noticed a small typo in an error message, so this PR fixes it. `s/inconsisntencies/inconsistencies/`

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [x] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
Thanks for the fix. Hopefully you didn't see that error during usage! Luckily the next version eliminates this style of index overwriting entirely, so it should never leave on inconsistent state even if cancelled halfway through doing something.

PR #593, opened 2020-12-29
Replace Link with Snapshot (WIP)

cdvv7788:
# Summary
Use Snapshot everywhere, instead of the old Link schema. This is a work in progress, and most likely unstable at this point.

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [X] Internal architecture


cdvv7788:
Extractors are working using snapshot only. Pending:
- HTML index is disabled at the moment. Refactor pending
- All of the parsers (except the base txt one) are still using link
- The generated index detail may be inconsistent with previous versions. Need to check this in detail before proceeding
- Only `add` and `init` commands tested so far. Need to work on the other commands.

cdvv7788:
Most commands are functional now. There are issues with what some of the return, but they don't raise any issue when running them:
- Add is still using parsers that require links
- Init is not correctly recreating the folders that are not in the main index but are in the filesystem
- Status is not displaying the right status for everything
- Server is returning a 500 error (with nothing being shown in the console)

Fixing these issues is what I am working on at the moment.

cdvv7788:
The most obvious issues are fixed now. All of the commands are partially functional. Now I will focus on:
- Getting the tests running
- Restore the html-related functionalities
- Find a good way to handle tags with the new structure
- Make sure that the index is correct when compared with the previous version

archivebox/parsers/pocket_api.py
```diff
@@ -96,7 +99,7 @@ def should_parse_as_pocket_api(text: str) -> bool:
 
 
 @enforce_types
-def parse_pocket_api_export(input_buffer: IO[str], **_kwargs) -> Iterable[Link]:
+def parse_pocket_api_export(input_buffer: IO[str], **_kwargs) -> Iterable[Model]:
```
pirate, line 102: not `Iterable[Snapshot]`?

cdvv7788: Cannot do that. Django will not let me import `Snapshot` directly, even with the typing condition. This was the closest I could think of.

pirate: ok

pirate:
Whoops closing this was not intentional, it auto-closed when v0.5.0 was deleted after merging.

Issue #597, opened 2021-01-03
Bugfix: sqlite3.IntegrityError: NOT NULL constraint failed: core_archiveresult.cmd_version and .output


coisnepe:
#### Describe the bug
Cannot start archivebox after moving from `nikisweeting/archivebox` to `archivebox/archivebox`. Django/Sqlite error ensues.

#### Steps to reproduce
Might be related to #596 ? I killed my docker container running `nikisweeting/archivebox:latest` and tried recreating a container with `archivebox/archivebox:latest`, linking to the same old volume. I did prepend the server command with `archivebox init` in order to run 3 new migrations. The container fails to start, issuing the following error.

Looks like [this is the migration that not going through](https://github.com/ArchiveBox/ArchiveBox/blob/31d5087b93d133a0a881a80e9695410bdfe87298/archivebox/core/migrations/0007_archiveresult.py), which might point at a corrupted/faulty json index file?

#### Screenshots or log output

```
[*] Verifying main SQL index and running migrations...
    √ /data/index.sqlite3

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/sqlite3/base.py", line 413, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.IntegrityError: NOT NULL constraint failed: core_archiveresult.cmd_version

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 227, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/migration.py", line 124, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/operations/special.py", line 190, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/app/archivebox/core/migrations/0007_archiveresult.py", line 33, in forwards_func
    ArchiveResult.objects.create(extractor=extractor, snapshot=snapshot, cmd=result["cmd"], cmd_version=result["cmd_version"], 
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
[...]

PR #599, opened 2021-01-04
fix: Give cmd_version a default value in case it is not present

cdvv7788:
# Summary

Give the `cmd_version` field a default value to handle the situation where it is not set in the origin file.

# Related issues

#597 
# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


archivebox/core/models.py
```diff
@@ -181,7 +181,7 @@ class ArchiveResult(models.Model):
     snapshot = models.ForeignKey(Snapshot, on_delete=models.CASCADE)
     cmd = JSONField()
     pwd = models.CharField(max_length=256)
-    cmd_version = models.CharField(max_length=32)
+    cmd_version = models.CharField(max_length=32, default="")
```
pirate, line 184:
```suggestion
    cmd_version = models.CharField(max_length=32, null=True, blank=True, default=None)
```
I'd rather not store emptystrings if possible.

cdvv7788: Ok, changed.

Issue #602, opened 2021-01-06
Bugfix: Docker container fails due to wrong chromium path


bash-stack:
had to execute 

`docker exec -it archivebox_archivebox_1 "ln -s /usr/bin/chromium /usr/bin/chromium-browser" `

pirate:
Can you post the exact error output you saw and the output of running `docker run archivebox/archivebox version`.

bash-stack:
sorry , i cannot remember exactly , it was the usual message when there is a binary missing in PATH ( Command not found) 

( it happened on aarch64 docker version)


 ```
√  ARCHIVEBOX_BINARY      /usr/local/bin/archivebox                                                    v0.5.0          valid                                                                                                                 
 √  PYTHON_BINARY          /usr/local/bin/python3.9                                                     v3.9.0          valid                                                                                                                 
 √  DJANGO_BINARY          /usr/local/lib/python3.9/site-packages/django/bin/django-admin.py            v3.1.3          valid                                                                                                                 
 √  CURL_BINARY            /usr/bin/curl                                                                v7.64.0         valid                                                                                                                 
 √  WGET_BINARY            /usr/bin/wget                                                                v1.20.1         valid                                                                                                                 
 √  NODE_BINARY            /usr/bin/node                                                                v15.3.0         valid                                                                                                                 
 √  SINGLEFILE_BINARY      /node/node_modules/single-file/cli/single-file                               v0.1.14         valid                                                                                                                 
 √  READABILITY_BINARY     /node/node_modules/readability-extractor/readability-extractor               v0.1.0          valid                                                                                                                 
 √  MERCURY_BINARY         /node/node_modules/@postlight/mercury-parser/cli.js                          v1.0.0          valid                                                                                                                 
 √  GIT_BINARY             /usr/bin/git                                                                 v2.20.1         valid                                                                                                                 
 -  YOUTUBEDL_BINARY       /usr/local/bin/youtube-dl                                                    -               disabled                                                                                                              
 √  CHROME_BINARY          /usr/bin/chromium                                                            v83.0.4103.116  valid                
```

pirate:
We need the command you ran and the exact error output to reproduce it. (also don't redact the output of `archivebox version`, the whole output it important for us to be able to reproduce bug reports)

bash-stack:
it is the following (aarch64) docker container:

`archivebox/archivebox   latest    fed7c56b810d   4 weeks ago    1.08GB`

[...]

PR #603, opened 2021-01-06
fix missing chromium path #602

bash-stack:
# Summary

fixes chrome binary in wrong path 

# Related issues

 #602

# Changes these areas

- [x] Bugfixes


pirate:
Why is this needed? It should already pick up `chromium` just fine: https://github.com/ArchiveBox/ArchiveBox/blob/dev/archivebox/config.py#L621

If you saw some error that prompted this can you post the exact error output and the output of `archivebox version`.

PR #612, opened 2021-01-10
Publish tag to Docker Hub

mAAdhaTTah:
This tags the versions released on Docker Hub so we
can rely on those versions rather than the sha.

# Summary

After 0.5.3 was published, I went to Docker Hub looking for an 0.5.3 tag so I could lock my containers to that
version but didn't find one. After poking around the GitHub workflow, I think I've got a solution that will
correctly publish a new version but this is hard to test. Not sure if you have a suggestion there besides
pushing a forked version (which I could do) or just... merging & seeing if it works and testing on the next tag?

# Related issues

N/A

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -52,15 +52,24 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-buildx-
       
-      - name: Get publish tag
-        id: publish
+      - name: Prepare tags to publish
+        id: prep
         run: |
+          # Always publish to latest.
+          TAGS="${{ secrets.DOCKER_USERNAME }}/archivebox:latest,archivebox/archivebox:latest"
           if [[ $GITHUB_REF == refs/tags/* ]]; then
-            TAG="${GITHUB_REF#refs/tags/}"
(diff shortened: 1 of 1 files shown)


pirate:
Could you add the short version too if possible:

```bash
VERSION="$(jq -r '.version' < "$REPO_DIR/package.json")"
SHORT_VERSION="$(echo "$VERSION" | perl -pe 's/(\d+)\.(\d+)\.(\d+)/$1.$2/g')"

# e.g.
# VERSION=0.5.3
# SHORT_VERSION=0.5
```

mAAdhaTTah:
@pirate I'm following there example here: https://github.com/docker/build-push-action/pull/119/files which includes publishing the three version types I've included in this PR. If you don't want this many publishes, I can slim this down, but using the bash script to generate the tags in general gives us more control over what we're publishing.

We could also shorten the version sha we're publishing too if we wanted. Let me know what you think of this so far.

pirate (approved):

PR #614, opened 2021-01-11
Link removal

cdvv7788:
# Summary
Reopening https://github.com/ArchiveBox/ArchiveBox/pull/593 with `dev` as base. The current focus is:
- Getting the tests running (11 out of 46 are failing at the moment)
- Restore the html-related functionalities
- Find a good way to handle tags with the new structure
- Make sure that the index is correct when compared with the previous version

# Related issues

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [X] Internal architecture


archivebox/themes/default/link_details.html
```diff
@@ -233,7 +233,7 @@
                 }
             }
         </style>
-        <link rel="stylesheet" href="../../static/bootstrap.min.css">
+        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/css/bootstrap.min.css">
```
pirate, line 236: can you change this back, it has to all be self-hosted so that it can operate 100% offline

cdvv7788: @pirate I was going to ask about this. The js is actually using the cdn, so I will need to change that too. The static indexes (outside of django) cannot find these files. Does the static folder needs to be copied to the archive and I broke it? I am almost done fixing the tests (just one missing). I am starting to check against the current live version.

pirate:
@cdvv7788 what's left on this PR, maybe we can have a quick call to hand it off and then I can finish it?

cdvv7788:
There are several things missing. The tests are mostly passing, but the html, specially the functionalities related to the static files need some review. Also, checking the index integrity and adding back the tags functionalities is pending.

pirate:
fyi I merged `dev` into this for you and fixed all the conflicts: https://github.com/ArchiveBox/ArchiveBox/pull/638

don't want it getting too far behind

cdvv7788:
Got it, thanks!. I am a little busy right now, but maybe next week I can find some time to finish it. It is not that far from finished.

pirate:
Closing this in favor of #638

PR #623, opened 2021-01-20
Fix dependency dict entries

thedanbob:
# Summary

This fixes a warning about the Node dependency being missing when `SAVE_SINGLEFILE=False` as well as a couple of other incorrect entries in the dependency dict.

# Related issues

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
weird, thanks, I don't know how those mistakes were committed originally. maybe a merge resolution gone wrong when I reordered them

PR #625, opened 2021-01-20
Update sonic.py

jdcaballerov:
Sonic buffer accepts 20.000 bytes not unicode characters, since the chunking here is on unicode characters, sending 20.000 characters will overflow sonic's buffer.
UTF-8 can take up to 6 bytes, so sending less than (20.000 / 6) rounded minus should be ok.

# Summary

# Related issues

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


tabashir:
This patch worked to prevent the error on my system. Are there any issues or side effects expected to look out for?

pirate:
Nope, it's a harmless change afaik @tabashir

Issue #617, opened 2021-01-15
Question: archivebox server throws 500 error


winteriscariot:
I'm using nginx in front of archivebox on my local intranet to access my archives, and it's throwing a 500 error whenever I simply load the front page. nginx is recording no errors, so I assume this is happening internal to archivebox. The `archivebox server` console -- running in a screen -- simply reports the 500 error and nothing further.

Is there anything I can do to troubleshoot this? Perhaps increase the verbosity of `archivebox server` at the console so I can see where the fault exists?

tried an `archivebox init` on my archive directory to no result

I can dump the archivebox list to html, which is what I'm doing for now, but it's not ideal since I can change the order of links (I prefer newest links at the top of the list, while the html dump puts oldest at top; it's a pet peeve but the archivebox server allows me to change that).

Arch Linux, archivebox 0.5.3 via pip

Thanks!

winteriscariot:
Here is the result of running `archivebox server --debug` and then loading (at which point I receive the 500 error):

```
$ archivebox server --debug
[i] [2021-01-15 16:31:07] ArchiveBox v0.5.3: archivebox server --debug
    > /mnt/storage/Archive

[+] Starting ArchiveBox webserver...
    Hint: The admin username is winteriscariot

Performing system checks...

System check identified no issues (0 silenced).
January 15, 2021 - 16:31:07
Django version 3.1.3, using settings 'core.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
[15/Jan/2021 16:31:12] "GET /admin/core/snapshot/ HTTP/1.0" 500 145
[15/Jan/2021 16:31:17] "GET / HTTP/1.0" 302 0
[15/Jan/2021 16:31:17] "GET /admin/core/snapshot/ HTTP/1.0" 500 145
```

winteriscariot:
After creating a new archive dir without the existing archives, I was able to avoid the 500 error. Therefore I'm marking this as closed, since it must be something to do with one of my links.

mAAdhaTTah:
If you set DEBUG=True in your config, you'll get a Django debug stack trace, which can help with stuff like this.

winteriscariot:
Hm the issue has started to recur. thanks for that @mAAdhaTTah -- i was able to get more info when loading /admin/core/snapshot/:

```
Environment:


Request Method: GET
Request URL: http://searx.local/admin/core/snapshot/

Django Version: 3.1.3
[...]

PR #632, opened 2021-01-24
Fix issue #617 by using mark_safe in combination with format_html

aggroskater:
I have no experience with Django, so all I'm really going off of is this
stackoverflow

https://stackoverflow.com/a/64498319

which cited this bit of Django documentation:

https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.html.format_html

After using this method, I no longer get the 500 error or KeyError
exception, and can browse the local server and interact with the single
entry in it (the problematic URL in ArchiveBox#617 with curly braces).

Whether this is the "right" method or not, I have no idea. But it is at
least a start.

# Summary

Fixes 500 errors on /public/ URL, experienced in issue #617. May not completely resolve all issues in #617.

# Related issues

None.

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/index/html.py
+++ b/archivebox/index/html.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from typing import List, Optional, Iterator, Mapping
 from pathlib import Path
 
-from django.utils.html import format_html
+from django.utils.html import format_html, mark_safe
 from collections import defaultdict
 
 from .schema import Link
@@ -161,4 +161,4 @@ def snapshot_icons(snapshot) -> str:
             output += '<a href="{}" class="exists-{}" title="{}">{}</a> '.format(canon["archive_org_path"], str(exists),
                                                                                         "archive_org", icons.get("archive_org", "?"))
 
-    return format_html(f'<span class="files-icons" style="font-size: 1.1em; opacity: 0.8">{output}<span>')
+    return format_html('<span class="files-icons" style="font-size: 1.1em; opacity: 0.8">{}<span>', mark_safe(output))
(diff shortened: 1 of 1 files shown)


archivebox/index/html.py
```diff
@@ -161,4 +161,4 @@ def snapshot_icons(snapshot) -> str:
             output += '<a href="{}" class="exists-{}" title="{}">{}</a> '.format(canon["archive_org_path"], str(exists),
                                                                                         "archive_org", icons.get("archive_org", "?"))
 
-    return format_html(f'<span class="files-icons" style="font-size: 1.1em; opacity: 0.8">{output}<span>')
+    return format_html('<span class="files-icons" style="font-size: 1.1em; opacity: 0.8">{}<span>', mark_safe(output))
```
pirate, line 164: I think doing `mark_safe(output)` at any point would be dangerous because it allows untrusted text to slip by the XSS escaping process. (`format_html` is used to compose html fragments while still escaping them to avoid XSS) If an attacker put `'><script>onload('haxxed')</script` in their url's filename, it might render here and pwn our user.

I think providing them as separate args is enough 🤞, does this work instead?
```suggestion
    return format_html('<span class="files-icons" style="font-size: 1.1em; opacity: 0.8">{}<span>', output)
```

aggroskater: Yep. Confirmed it still works for me.

PR #655, opened 2021-02-16
v0.6.0 Release

pirate:
new features: 

- new ArchiveResult admin UI, with full editing ability + inline list under each Snapshot admin page
- add `init --quick` and `server --quick-init` options to quickly update the db version without doing a full re-init (for users with large archive collections this will make version upgrades a lot faster / less painful)
- new `SNAPSHOTS_PER_PAGE=40` and `MEDIA_MAX_SIZE=750m` config options
- log all errors with full tracebacks to new `data/logs/errors.log` file (so users no longer have to run in --debug mode to see error details)
- allow hotlinking directly to specific extractor output on the snapshot detail page using URL `#hash` e.g. `/archive/<timestamp>/index.html#git`
- add ability to view snapshot matching a given URLs by visiting `/archive/https://example.com/some/url` -> redirects to -> `/archive/<timestamp>/index.html` (also works without scheme `/archive/example.com`)
- add Django Debug Toolbar + `djdt_flamegraph` for developers to profile UI performance
- new `archivebox add --tag=tag1,tag2,tag3 ...` option and Add page UI form field for tagging new links

enhancements:

- lots of performance improvements! (in testing, the main index was brought down from 10-14 second load times to 114ms once cache warms up)
- speed up main codebae hotspots `snapshot_icons`, `latest_title`, `archive_size` using django cache
- integrity and correctness improvements to readability, mercury, warc, and other extractors
- lots of small UI improvements and CLI improvements

bugfixes:

- fix stdin/stdout/stderr handling for some edge cases in Docker/Docker-Compose
- lots of minor template fixes


pirate:
This PR is only partial and is is continued here: https://github.com/ArchiveBox/ArchiveBox/pull/680

Issue #578, opened 2020-12-10
Feature Request: Scheduling Archival from the UI


BlipRanger:
## Type 

 - [ ] General question or discussion
 - [x] Propose a brand new feature
 - [ ] Request modification of existing behavior or design

## What is the problem that your feature request solves

Currently scheduling ingestion of new urls requires writing a cron job external to the web UI (external to the docker container in my case) which isn't entirely ideal in a docker/self-contained setup. I believe this would be a nice convenience feature for users that might want to manage the entire operation of AB from within the web UI. 
 
## Describe the ideal specific solution you'd want, and whether it fits into any broader scope of changes

This feature would add a method for setting up scheduled pulls from various data sources via the web UI rather than only externally via cron. I specifically imagine at least a way to specify a RSS feed to be subscribed to that it can watch for new content from (something like Wallabag in my particular imagined use case). Technically I think this would involve a new menu/button in the UI and should dovetail with the internal scheduling processes already available. 

## How badly do you want this new feature?

 - [ ] It's an urgent deal-breaker, I can't live without it
 - [x] It's important to add it in the near-mid term future
 - [ ] It would be nice to have eventually

---

 - [x] I'm willing to contribute dev time / money to fix this issue
 - [x] I like ArchiveBox so far / would recommend it to a friend
 - [ ] I've had a lot of difficulty getting ArchiveBox set up

pirate:
Yeah this is definitely on our mind, it probably won't be added for a couple versions but this is definitely something I've been planning.

It's blocked by adding a background queue system like Huey: https://github.com/ArchiveBox/ArchiveBox/issues/91

In the meantime I recommend using docker-compose instead of docker alone, as it allows you to declaratively define your scheduled imports all in one place (you can see the `docker-compose.yml` commented out section for an example of how to do that).

BlipRanger:
Gotcha, I saw the future queuing system and that makes sense! And yes, currently using compose, so I'll look into doing that. Thanks!

pirate:
Here's my proposed implementation of a new model to track scheduled imports:

```python3
class ScheduledTask(models.Models):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    
    name = models.CharField(max_length=128, required=True, unique=True)
    schedule = models.CharField(max_length=32, default='weekly')
    enabled = models.BooleanField(default=True)

    added = models.DateTimeField(auto_now_add=True)
[...]

PR #707, opened 2021-04-16
#578: Add ability to schedule and manage recurring imports via the admin UI

pirate:
Fixes: #578

**Remaining TODOs:**

- [ ] figure out which python scheduler to use
  - huey (my current favorite)
  - celery (ugh...)
  - APScheduler (will require lots of manual models and concurrency control code)
  - yacron (not sure if it can be configured dynamically)
  - dramatiq (doesn't support sqlite)
- [ ] decide whether to continue supporting system crontab at all, or tear it out (imo we should just tear it out and move to using an internal scheduler)
- [ ] fork the scheduled task worker off the server process automatically on startup, so no need to run separate `archivebox schedule --foreground` process manually
- [ ] figure out how to enforce "at least once" or "at most once" concurrency model for scheduled tasks

### DIFF
--- a/archivebox/main.py
+++ b/archivebox/main.py
@@ -7,7 +7,7 @@ import platform
 from pathlib import Path
 from datetime import date
 
-from typing import Dict, List, Optional, Iterable, IO, Union
+from typing import Dict, List, Tuple, Optional, Iterable, IO, Union
 from crontab import CronTab, CronSlices
 from django.db.models import QuerySet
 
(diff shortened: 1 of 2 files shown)


jhotujec:
Is this still alive? I'm looking for something similar, I would be ready to implement it if original author gave up on PR.

pirate:
This is still alive but very slow moving. If you need an immediate solution I recommend building a quick and dirty GUI to modify the scheduler cron config directly or call out to the `archivebox schedule` CLI.

GlassedSilver:
Any updates on this? IMHO this feature, even though it's mostly a UX improvement rather than a pure programmatic change could greatly improve ArchiveBox's utility.

pirate:
Going to close this PR as the new django-huey-monitor / Celery based system I'm adding uses a totally different approach to handle scheduling tasks, and much of the UI we get for free.

PR #741, opened 2021-05-07
fix: use 644 instead of 755 for the configuration file

srevinsaju:
# Summary

ArchiveBox attempted to create configuration files 'ArchiveBox.conf'
which had execute permissions '+x' for everyone, which didnt seem safe
which seems to introduce security vulnerabilities.

# Related issues
None

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture
- [x] Snapshot data layout on disk


pirate:
If you want to change this, you should change only the permissions of the config file, not the entire OUTPUT_PERMISSIONS, as `+x` is needed on directories in the archive output.

pirate:
Fixed in 8230f88 will be out with v0.6.3

Issue #788, opened 2021-07-14
Use `whitenoise` for serving static files instead of `django.views.static.serve`


pirate:
https://github.com/evansd/whitenoise

> With a couple of lines of config WhiteNoise allows your web app to serve its own static files, making it a self-contained unit that can be deployed anywhere without relying on nginx, Amazon S3 or any other external service. (Especially useful on Heroku, OpenShift and other PaaS providers.)

```python3
MIDDLEWARE = [
  # 'django.middleware.security.SecurityMiddleware',
  'whitenoise.middleware.WhiteNoiseMiddleware',
  # ...
]

STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```

Also has the added benefit of supporting Byte-Range requests (so video files will be seekable), brotli/gzip compression, and other nice headers

pirate:
I tried this and it doesn't actually help us with anything, since we need byte-range requests for user-created archive content, not the staticfiles.

Removed it in favor of configuring uwsgi to support honor byte-range requests instead.

PR #833, opened 2021-08-25
using whitenoise for serving static files

ViAsmit:
# Summary

This PR  adds whitenoise for serving static files

# Related issues

Fixes Issue #788 

# Changes these areas

- [] Bugfixes
- [] Feature behavior
- [] Command line interface
- [x] Configuration options
- [x] Internal architecture

### DIFF
--- a/archivebox/core/settings.py
+++ b/archivebox/core/settings.py
@@ -68,6 +68,15 @@ MIDDLEWARE = [
 
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
 
+# https://whitenoise.evans.io/en/stable/django.html
+# pip install whitenoise[brotli]
+WHITENOISE_INDEX_FILE = True
+WHITENOISE_USE_FINDERS = True
+WHITENOISE_ALLOW_ALL_ORIGINS = True
+
+# Use django-storages for user-uploaded files / archived content
+# https://django-storages.readthedocs.io/en/latest/
+
 AUTHENTICATION_BACKENDS = [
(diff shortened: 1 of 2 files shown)


pirate:
Thanks for this! It will involve quite a few more changes than just this PR, so it might take me a while to finish it off, but this is a good start.

Remaining things to check:
- [ ] add/verify dependency works in brew build
- [ ] add/verify dependency works in apt build
- [ ] add/verify dependency works in docker build 
- [ ] confirm static archive export urls work
- [ ] confirm headers served with staticfiles match previous headers
- [ ] check for performance changes / regressions

ViAsmit:
Yeah I am working on them. Will try to add more commitss in this PR .

archivebox/core/settings.py
```diff
@@ -65,10 +66,13 @@
     'core.middleware.CacheControlMiddleware',
 ]
 
+STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```
pirate, line 69:
```suggestion
STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'

# https://whitenoise.evans.io/en/stable/django.html
# pip install whitenoise[brotli]
WHITENOISE_INDEX_FILE = True
WHITENOISE_USE_FINDERS = True
WHITENOISE_ALLOW_ALL_ORIGINS = True

# Use django-storages for user-uploaded files / archived content
# https://django-storages.readthedocs.io/en/latest/
```

Issue #773, opened 2021-06-23
Feature Request: Support for external backend for authentication


dot-mike:
## Type 
 - [x] Propose a brand new feature

## What is the problem that your feature request solves

I want to authenticate with a custom backend like Authelia

## Describe the ideal specific solution you'd want, and whether it fits into any broader scope of changes
User should be able to switch to a custom backend in the settings for authentication and set custom header

## What hacks or alternative solutions have you tried to solve the problem?
Creating a middelware manually and setting proper header for authentication and adding it to `settings.py` solves this issue locally... see: https://docs.djangoproject.com/en/3.2/howto/auth-remote-user/

Authelia sets the `Remote-User` which translates to `HTTP_REMOTE_USER`

## How badly do you want this new feature?
 - [x] It would be nice to have eventually

---
 - [x] I like ArchiveBox so far / would recommend it to a friend

pirate:
I'd be open to merging PRs from other people to add this ability, but I'm unlikely to build it myself due to limited time.

If anyone wants to work on this please outline your proposed implementation here before you starting coding it!

pirate:
Closing is as complete for now because ArchiveBox supports both reverse proxy auth and LDAP now.

Open new issues for any future requests!

pirate:
Reopening to track adding the CSRF_TRUSTED_ORIGINS config option.

PR #866, opened 2021-09-30
Support for Reverse Proxy authentication backends (like authelia)

ajgon:
# Summary

Adds support to reverse proxy authentication backends (like authelia) via configured HTTP header. 

# Related issues

#773 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture


archivebox/core/middleware.py
```diff
@@ -35,3 +38,23 @@ def middleware(request):
...
     return middleware
+
+class ReverseProxyAuthMiddleware(RemoteUserMiddleware):
+    header = 'HTTP_{normalized}'.format(normalized=REVERSE_PROXY_USER_HEADER.replace('-', '_').upper())
+
+    def process_request(self, request):
+        if REVERSE_PROXY_WHITELIST == '':
+            return
+
+        ip = request.META.get('REMOTE_ADDR')
```
pirate, line 49: The name of this header should probably also be a config option. cloudflare, nginx, etc. different reverse proxies set different headers for the user's true origin ip address.

pirate:
Where is `REVERSE_PROXY_USER_HEADER` actually handled? I don't see any code passing it to some auth/auth skipping mechanism?

archivebox/core/middleware.py
ajgon: This is not the header used by the proxy. This the remote IP address of the caller which is used against whitelist. Only if the IP is included there, the header is taken into consideration. The header itself is set here: https://github.com/ArchiveBox/ArchiveBox/pull/866/files#diff-618efc77bbfa507509fde994d8fa689045cbf17527233bd4d2992ce18d8dcd26R43 and it's configurable via `REVERSE_PROXY_USER_HEADER` config parameter.

ajgon:
> Thanks for this PR and the documentation PR to go with it! It looks like it's on a good track so far.
> 
> Quick question: Where is `REVERSE_PROXY_USER_HEADER` actually handled? I don't see any code passing it to some auth mechanism?

Everything is handled by standard `RemoteUserMiddleware` class from django. I've just created subclass for it to handle customizable header and whitelisting. I was basing this on [this](https://docs.djangoproject.com/en/3.2/howto/auth-remote-user/).

ajgon:
Let me also explain how this is supposed to work, to ensure that we're on the same page :) :

- User visits archivebox, which is protected by auth proxy (let's use authelia for the example, but the pattern is the same for vouch, keycloak etc.)
- User signs in in authelia. Authelia takes care of authentication, and is our "source of truth" about user identity. The idea is, that we trust the proxy, it handled authentication properly and now wishes to let us know about it.
- Authelia sends login of authenticated user in `Remote-User` header (it also sends `Remote-Email`, and some other ones, but the one which we are interested in, is only user). This is the username of the user which we wish to authenticate archivebox side (so it has to match database records).
    - Other auth systems use different headers (`X-Remote-User` for example) - that's why, the header name is configurable via `REVERSE_PROXY_USER_HEADER` config option.
- Now, this is also a potential security flaw, as malicious attacker can send it's own `Remote-User: admin` header and try to log in, mimicing the proxy. Usually it shouldn't be the problem, because when using authelia, archivebox should never be exposed to the world, it should always be proxied through authelia.
- But we live in a real world, and misconfiguration accidents happen :( To add extra layer of security, I introduced a `REVERSE_PROXY_WHITELIST` mechanism. It's very simple concept - we only allow the given list of IP classes (CIDRs), to actually provide `Remote-User` header. Every other IP sending header is ignored. Usually, this should be configured only to authelia IP. For example, for docker environments setting this to `172.16.0.0/12` should usually be sufficient enough.
    - We are only interested in IP of the proxy itself, we should never care about the client IP (so, we should never use any `X-Forwarded-For` and similar headers). This is handled by `request.META.get('REMOTE_ADDR')` which is raw IP of the connecting service (in this case - the proxy).
- By default `REVERSE_PROXY_WHITELIST` is set to empty string, which means "blacklist everything". This effectively means, this feature is disabled by default - so if somebody doesn't care about and sets up archivebox "out from the box", they are not exposed to any security issues.

Hopefully this clears things up, and explain what I'm trying to build here :)

archivebox/config.py
```diff
@@ -82,17 +82,19 @@
...
+        'PUBLIC_INDEX':              {'type': bool,  'default': True},
+        'PUBLIC_SNAPSHOTS':          {'type': bool,  'default': True},
+        'PUBLIC_ADD_VIEW':           {'type': bool,  'default': False},
+        'FOOTER_INFO':               {'type': str,   'default': 'Content is hosted for personal archiving purposes only.  Contact server owner for any takedown requests.'},
+        'SNAPSHOTS_PER_PAGE':        {'type': int,   'default': 40},
+        'CUSTOM_TEMPLATES_DIR':      {'type': str,   'default': None},
+        'TIME_ZONE':                 {'type': str,   'default': 'UTC'},
+        'REVERSE_PROXY_USER_HEADER': {'type': str,   'default': 'Remote-User'},
+        'REVERSE_PROXY_WHITELIST':   {'type': str,   'default': ''},
     },
```
dugite-code, line 98: To allow for logout flows in some SSO providers like authentik having the LOGOUT_REDIRECT_URL configurable for this PR would be useful.

```suggestion
        'LOGOUT_REDIRECT_URL':   {'type': str,   'default': '/'},
    },
```

Note: would also need line 37 in settings.py changed
https://github.com/ArchiveBox/ArchiveBox/blob/95cf85f8cf4a036ebf2f332fe29d0735219cf715/archivebox/core/settings.py#L37

```
LOGOUT_REDIRECT_URL = os.environ.get("LOGOUT_REDIRECT_URL")
```

iarp:
@pirate @dugite-code Hi, any chance of this PR being merged soon? Or where are things left? It would be quite nice to get it through.

dugite-code:
I've been running this patched into the current version of ArchiveBox for a while. As long as the user exists initially the SSO works as expected, however any auto-created users do not have the correct permissions assigned leaving users stuck on the login page with the message: `You are authenticated as example.user, but are not authorized to access this page. Would you like to login to a different account?` They can manually navigate to the public url and show as logged in.

The user needs to be set as staff and at least given viewing permissions, although it appears you can still add/remove snapshots with just viewing permissions.

iarp:
What about adding a server config option that adds the `RemoteUserBackend` to `AUTHENTICATION_BACKENDS` and adding a note to the config page with something along the lines of "If you enable this, any users using this option with ArchiveBox must be set with correct permissions of ___".

Another option could be to alter RemoteUserBackend.[configure_user](https://github.com/django/django/blob/main/django/contrib/auth/backends.py#L238) to add the necessary permissions and flags (depending on more server config options).

pirate (approved):

pirate:
This is now documented here: https://github.com/ArchiveBox/ArchiveBox/wiki/Configuration#reverse_proxy_user_header

ajvpot:
> I've been running this patched into the current version of ArchiveBox for a while. As long as the user exists initially the SSO works as expected, however any auto-created users do not have the correct permissions assigned leaving users stuck on the login page with the message: `You are authenticated as example.user, but are not authorized to access this page. Would you like to login to a different account?` They can manually navigate to the public url and show as logged in.
> 
> The user needs to be set as staff and at least given viewing permissions, although it appears you can still add/remove snapshots with just viewing permissions. Note: you cannot modify auto created users after the fact.

I'm also experiencing this. The experience is broken. Would it be possible to add another config option for which permissions to grant automatically created users?

pirate:
> Would it be possible to add another config option for which permissions to grant automatically created users?

Yes, we recently added this same fix for LDAP auth here: https://github.com/ArchiveBox/ArchiveBox/pull/1335

I would just set `is_staff=True` and `is_superuser=True` for now, don't bother with more granular permissions as the rest of the codebase doesn't support row-level permissions yet.

Unfortunately I'm a bit overloaded with paying client work right now so I probably won't get around to implementing this myself, but if you submit a PR to add this I'd be happy to review it!

a10kiloham:
I'm getting CSRF errors when I try and do things like a re-crawl. I have my  From looking online it seems like there ought to be a setting to allow TRUSTED_HOSTS or otherwise USE_X_FORWARDED_HOST = True should be true in Django I think?
I have my settings as follows:
- REVERSE_PROXY_WHITELIST=172.0.0.0/8
- REVERSE_PROXY_USER_HEADER=Remote-User
which I believe should be correct? This is using Authelia and Caddy on top of the Nginx internally for cert management fwiw.

pirate:
It's possible, can you test if that change fixes it and report back?

- find the archivebox source code location on your machine `archivebox version | grep PACKAGE_DIR`
- then edit `PACKAGE_DIR/core/settings.py` and add `USE_X_FORWARDED_HOST = True`
- restart archivebox to see if that works

lkubb:
> I'm getting CSRF errors

Seeing the same during a regular login with the `dev` docker container. I could fix it with the following in `settings.py` (no `USE_X_FORWARDED_HOST` necessary)

```py
CSRF_TRUSTED_ORIGINS = ["https://my.archivebox.domain"]
```

Side note: The logout button does not work either since it results in a GET request, but needs a POST with CSRF token.

a10kiloham:
As a quick fix for this I just edited my docker-compose as follows (inserting my reverse proxied external address into the trusted origins line.
```
services:
    archivebox:
      build:
        context: .
        dockerfile_inline: |
          FROM archivebox/archivebox:dev
          RUN echo 'CSRF_TRUSTED_ORIGINS = ["https://archive.example.com"]' >> /app/archivebox/core/settings.py
      command: server --quick-init 0.0.0.0:8000
      container_name: archivebox
```

pirate:
Ok this should be fixed now on `dev`: 9c35f3dd

I've added `CSRF_TRUSTED_ORIGINS` to the available config, no need for the dockerfile override hack anymore.

PR #908, opened 2021-12-20
Fix #725 - correctly parse tags on json import

### LINKED ISSUE #725: Pocket and Pinboard imports causing tags to be split incorrectly into individual characters w/ broken hyphenation
A simple bug due to a `.split()` or `set()` somewhere on the tags_str instead of the tags list. Should be easy to fix.

image

We should also add a filter to prevent emptystring / whitespace-only tags:
image

hannah98:
# Summary

This PR fixes #725 - instead of tags being split on each character, this fix mimics a code block further down in the same file to create a list of tags first before calling the ```save_tags()``` method.

https://github.com/ArchiveBox/ArchiveBox/blob/32764347ce2e59919f763c552bd3e250f49c2f5b/archivebox/index/sql.py#L107-L109 

# Related issues

#725 

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/index/sql.py
+++ b/archivebox/index/sql.py
@@ -34,8 +34,11 @@ def write_link_to_sql_index(link: Link):
     from core.models import Snapshot, ArchiveResult
     info = {k: v for k, v in link._asdict().items() if k in Snapshot.keys}
     tags = info.pop("tags")
-    if tags is None:
-        tags = []
+
+    tag_set = (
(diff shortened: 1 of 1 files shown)


pirate:
Thanks @hannah98 !!!

PR #911, opened 2021-12-30
Added TAG_SEPARATOR_PATTERN option for splitting tags

hannah98:
# Summary
This PR adds the ```TAG_SEPARATOR_PATTERN``` option.  This option is a regex character class that will be used when splitting the tag string.
It defaults to ```[,]``` which will split tags on commas.  Splitting tags is currently done on commas so this will be backwards compatible and should not introduce breaking changes.

Some Examples:
* Splitting on spaces: ```TAG_SEPARATOR_PATTERN="[ ]"```
* Splitting on spaces and commas: ```TAG_SEPARATOR_PATTERN="[ ,]"```
* Splitting on spaces, commas, and semicolons: ```TAG_SEPARATOR_PATTERN="[ ,;]"```

---

* All previously passing tests still pass.
* I can give the Wiki text so that the Wiki page for the configuration can be updated with this new option.

# Related issues

#725 

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/config.py
+++ b/archivebox/config.py
@@ -79,7 +79,7 @@ CONFIG_SCHEMA: Dict[str, ConfigDefaultDict] = {
         'URL_BLACKLIST':            {'type': str,   'default': r'\.(css|js|otf|ttf|woff|woff2|gstatic\.com|googleapis\.com/css)(\?.*)?$'},  # to avoid downloading code assets as their own pages
         'URL_WHITELIST':            {'type': str,   'default': None},
         'ENFORCE_ATOMIC_WRITES':    {'type': bool,  'default': True},
-        'TAG_SEPARATOR_PATTERN':    {'type': str,   'default': '[,]'},
+        'TAG_SEPARATOR_PATTERN':    {'type': str,   'default': r'[,]'},
     },
 
     'SERVER_CONFIG': {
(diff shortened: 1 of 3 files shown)


pirate:
This looks awesome, thanks for working on it!

One small note: If it's a REGEX we should probably change the name from `TAG_SEPARATORS` which implies list (usually a CSV/JSON string in configs), to something like `TAG_SEPARATOR_PATTERN` which implies a REGEX.

hannah98:
Understood.  I have updated the code to rename TAG_SEPARATORS to TAG_SEPARATOR_PATTERN.

archivebox/config.py
```diff
@@ -79,6 +79,7 @@
         'URL_BLACKLIST':            {'type': str,   'default': r'\.(css|js|otf|ttf|woff|woff2|gstatic\.com|googleapis\.com/css)(\?.*)?$'},  # to avoid downloading code assets as their own pages
         'URL_WHITELIST':            {'type': str,   'default': None},
         'ENFORCE_ATOMIC_WRITES':    {'type': bool,  'default': True},
+        'TAG_SEPARATOR_PATTERN':    {'type': str,   'default': '[,]'},
```
pirate, line 82:
```suggestion
        'TAG_SEPARATOR_PATTERN':    {'type': str,   'default': r'[,]'},
```

rmohns:
Wiki does not appear to have been updated with instructions on how to use this. @hannah98 or @pirate ?

hugochisholm:
I'm also having a bit of trouble with implementing this in a docker-compose use case. Should I be able to just add `TAG_SEPARATOR_PATTERN=r'[ ,;]'` to the environment variables in the docker-compose.yml?  

I'm having no luck if that's the case.

pirate:
Oops You're right I haven't updated it yet, the configuration page is a bit behind. It used to be publicly editable but some people abused it by adding virus links all over, so I had to lock it down. Unfortunately I don't know if there's a way to accept PRs for the wiki?

If anyone wants to contribute to wiki, I can just copy paste suggestions in from comments here and commit them myself.

Issue #916, opened 2022-01-24
Feature Request: Consider providing an official AppImage


noctux:
## Type 

 - [ ] General question or discussion
 - [X] Propose a brand new feature
 - [ ] Request modification of existing behavior or design

## What is the problem that your feature request solves

ArchiveBox as a software bundles multiple "moving pieces", including a Django website, several node-js based services, youtube-dl, etc.pp. This makes maintaining an installation, especially on fast moving rolling-release distributions a time-consuming task.
The proposed solution to this is running the docker container, which among other things requires root access (so it does not work e.g. on my universities shell account) and a complicated setup.
It would be nice if there was a simpler way to maintain a personal archivebox, while, like the Readme states, "avoid polluting your host system with extra dependencies".

## Describe the ideal specific solution you'd want, and whether it fits into any broader scope of changes

[AppImages](https://appimage.org/) allow to pack an application, along with its dependencies, into a single file binary. This binary contains a squashfs-based filesystem image of your application which will be mounted and then used to run your app.
Thus, distributing is (ideally) as simple as downloading the binary from the official release page, setting it executable, and be ready to run.

### Cons:
- Another maintanence burden
- AppImages do not feature isolation features, in contrast to docker. This can be both good and bad:
  + The appimage could for instance allow to run an host-provided chromium/ffmpeg/youtube-dl, can be interacted with by other means than the webinterface, ...
  + Several deployment strategies (such as a systemd-unit activated by an systemd-timer) do not work with docker at all, furthermore, these usually provide their own hardening options that allow security isolation

### Pros:
- More or less portable, single file binary might help adoption
- Similar to the existing docker container, can probably be integrated into the existing release automation I've seen on master

## How badly do you want this new feature?

 - [ ] It's an urgent deal-breaker, I can't live without it
 - [ ] It's important to add it in the near-mid term future
 - [X] It would be nice to have eventually

---

 - [X] I'm willing to contribute [dev time](https://github.com/ArchiveBox/ArchiveBox#archivebox-development) to fix this issue

I've toyed around with that idea last weekend and used [pkg2appimage](https://github.com/AppImage/pkg2appimage) (in version pkg2appimage-1807-x86_64.AppImage) to create an appimage for archivebox:

```yaml
# Note: due to reasons, this requires a python3 (matching, e.g. python3.9 on both), python3-venv, nodejs and npm ON THE BUILDING HOST
app: archivebox
ingredients:
  dist: focal
  sources:
[...]

PR #922, opened 2022-02-03
WIP: Add AppImage workflow

noctux:
# Summary

As discussed in #916, I see value in a release of ArchiveBox as an AppImage along with the other releases.
Therefore, this commit adds a github workflow for automatically building an AppImage on pushes.

Please consider this PR as work in progress. The AppImage building works, but I'm not 100% sure as to how you wish to exactly integrate this workflow into the repository and how uploading to the releases page works. Currently, the AppImage is uploaded as an artifact of the CI-run to show that working artifacts are produced. Currently, you seem to have factored out all distribution-methods into their own submodules. Should I do the same for the appimage_dist?

Furthermore, I plan to write a short bit of documentation on the specifics of the AppImage (i.e.: what is packaged, what is not, that there is no sandboxing and that other distribution methods are preferred and potentially better supported). I'd probably put that into `docs/Install.md`. Does that sound like the correct location?

Any other comments/requests/*?

Thanks for your efforts!

# Related issues

#916

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -1,0 +1,76 @@
+name: Build AppImage
+
+on:
+  workflow_dispatch:
+  push:
+
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 1
+
+      - name: Install debian dependencies
+        run: |
+          sudo apt-get install -y \
+            python3.9             \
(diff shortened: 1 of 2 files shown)


pirate:
Unfortunately I don't have the bandwidth to support another release channel myself, but if you want to create a separate repo that releases ArchiveBox as an appimage, I can link to it from our README.

PR #927, opened 2022-02-12
Assigns a random User Agent to each request if no agent is defined.

matiaszanolli:
# Summary

Basically the title explains it all. It fetches a list of the most common desktop user agents when ArchiveBox starts up and every day after, from a Python requests plugin dedicated to mostly the same task (https://github.com/DavidWittman/requests-random-user-agent). 

I simplified the logic as much as possible so it fits as a simple class on the util module, and made the required change on each request. I also added a default fallback on the settings to the traditional behaviour, so if you define a custom CURL_USER_AGENT instead of leaving the field blank, it will use that agent instead of a random one.

It's still on testing phase, but I must say it works like a charm on my local installation.

# Changes these areas

- [x] Feature behavior
- [x] Configuration options


archivebox/main.py
```diff
@@ -280,7 +281,7 @@ def init(force: bool=False, quick: bool=False, setup: bool=False, out_dir: Path=
         stderr("[!] This folder contains a JSON index. It is deprecated, and will no longer be kept up to date automatically.", color="lightyellow")
         stderr("    You can run `archivebox list --json --with-headers > static_index.json` to manually generate it.", color="lightyellow")
 
-    existing_index = (out_dir / SQL_INDEX_FILENAME).exists()
+    existing_index = True  # (out_dir / SQL_INDEX_FILENAME).exists()
```
varenc, line 284: ?

sonic.cfg
```diff
@@ -0,0 +1,66 @@
+# Sonic
```
varenc, line 1: this is unrelated to user-agents?

archivebox/core/migrations/0021_auto_20220128_2125.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.1.14 on 2022-01-28 21:25
```
varenc, line 1: i don't understand how database migrations are related to user-agents

varenc:
I'm just a random user so don't pay me much heed but... doesn't this remove the friendly self-identifying part of the old user-agent? 

I'm talking about this part where it acts like a good bot and identifies itself:
`...ArchiveBox/{VERSION} (+https://github.com/ArchiveBox/ArchiveBox/)...`

I think it's good when bots property identify themselves, at least by default. IMHO I don't think that should be removed.

Other quick thoughts:
- This seems like A LOT of code to randomize User-Agents
- Lots of code seems unrelated
- There seems little need to download the list constantly.  If you do want to randomize it, a fixed hardcoded list, that's maybe updated every year or so, seems fine.
- To the above point, you can see that the list you are downloading daily [appears to only be updated a few times a year anyway](https://github.com/DavidWittman/requests-random-user-agent/commits/master/requests_random_user_agent/useragents.txt)
- Weird things might happen when the bot randomly switches between identifying itself as Linux Firefox, then as macOS Chrome, then as Windows Edge, etc.  Site behavior may change between requests and different archive outputs might not match each other.  Some of the user agents in that list are pretty obscure and might trigger odd behavior.  (Like identifying as `YaBrowser/21.11.4.727 Yowser`)
- If this is about anonymity, one IP address that rotates between browsers and operating systems constantly makes you pretty noticeable. 

If you're just trying to maximize anonymity, just set your archivebox's user-agent to a recent Windows 10 Chrome user-agent and make that your default.  That'll blend in better than some of the ones on that user agent list.  It could be the default for everyone, though to my point above, that would remove the good bot identifiers in the User-Agents, which I think should remain.

p.s. sorry i wrote so much.  I enjoyed reading your code since it gave me a nice overview of the codebase!

pirate:
I think this is a neat idea but I don't want to support it natively in ArchiveBox. This can be achieved with an external program that controls archivebox (e.g. with environment variables, python API calls, etc.) to mutate the agent on every snapshot.

I appreciate your work so far on it though @matiaszanolli and I'm sorry I didn't review this sooner. I encourage you to take this PR and turn it into an independent script that puppets ArchiveBox. :+1:

PR #967, opened 2022-04-19
Fix missing input redirection in a hint text

rossvor:
# Summary

Currently one of the hints after the setup seems to suggest that 'add' command supports newline separated URL text files directly, as far as I understand it doesn't support that (maybe it used to be able to do it?). Change hint to show that the shell input redirection is required for importing multiple URLs from a text file.

# Related issues

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
It does support files directly without the input redirection. :+1:

rossvor:
Gotcha, thanks. Doesn't seem to work for me for some reason. Will raise an issue.

PR #977, opened 2022-05-17
Fix Issue with Parenthesis in URL

RikkiSmith:
# SUmmary

This commit fix an issue when one user try to add an URL
containing an URL. The pattern matching stop at the open
parenthesis. The URL generated is wrong and it archives
a truncated one.

You can easily reproduce this behavior by adding any URL
to archive and see this will not correctly archived.

# Related issues

Did not find any related issue.

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/util.py
+++ b/archivebox/util.py
@@ -61,7 +61,7 @@ URL_REGEX = re.compile(
     r'(?:[a-zA-Z]|[0-9]'              # followed by allowed alphanum characters
     r'|[$-_@.&+]|[!*\(\),]'           #    or allowed symbols
     r'|(?:%[0-9a-fA-F][0-9a-fA-F]))'  #    or allowed unicode bytes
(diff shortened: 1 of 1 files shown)


aaronxn:
I'm looking forward to using this app. Please fix. If I knew more I would help.

pirate:
The reason we stop parsing at parens is to support loading URLs from markdown. A solution would have to detect if the content looks like markdown and parse it accordingly, which is difficult and why I haven't done it already.

pirate:
I'm going to close this for now because the markdown-parsing behavior is one of the most common uses of ArchiveBox. But don't worry, I still plan to implement markdown-detection in the future to resolve the underlying UX issue and allow parens in non-markdown URLs.

PR #996, opened 2022-07-05
Skip first line of the "JSON" file

barthalion:
ArchiveBox moves the file to parse to the sources directory and adds the
original filename at the top, making the file invalid.


pirate:
Can we add an `if ...: next(file)` condition in there to check for this specific case? e.g. 
```python
if peekfirstline(file).strip()[0] != "{":
    next(file)  # skip first line if it doesnt start with {
```

PR #1023, opened 2022-09-11
Added standalone dockerfile instructions

turian:
# Summary

`Dockerfile` now has standalone build instructions.

There are two motivations:
1) I immediately opened the `Dockerfile` so I could begin hacking. But the instructions therein aren't standalone, and standalone is nice.
2) I wasn't aware there was a "Developing Archivebox" doc. I assumed this would be in the docs, not in the README (which is pretty big).

# Related issues

n/a

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
Thanks for this!

Issue #1014, opened 2022-08-12
Bug: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb3 in position 264619: invalid start byte


turian:
#### Describe the bug

running archivebox on many URLs, it snagged with the following error:

```

[+] [2022-08-12 18:04:45] "www.ashra.com/news.php?m=A"
    http://www.ashra.com/news.php?m=A
    > ./archive/1660315483.000912
      > favicon
      > headers
      > singlefile
      > pdf
        Extractor timed out after 60s.
        Run to see full output:
            cd /data/archive/1660315483.000912;
            /usr/bin/chromium --headless --no-sandbox --disable-gpu --disable-dev-shm-usage --disable-software-rasterizer --run-all-compositor-stages-before-draw --hide-scrollbars --single-process --no-zygote "--user-agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/605.1.15 ArchiveBox/{VERSION} (+https://github.com/ArchiveBox/ArchiveBox/)" --window-size=1440,2000 --timeout=60000 --print-to-pdf http://www.ashra.com/news.php?m=A

      > screenshot
      > dom
      > wget
      > title
      > readability
      > mercury
      > media
    ! Failed to archive link: Exception: Exception in archive_methods.save_media(Link(url=http://www.ashra.com/news.php?m=A))

Traceback (most recent call last):
  File "/app/archivebox/extractors/__init__.py", line 109, in archive_link
    result = method_function(link=link, out_dir=out_dir)
  File "/app/archivebox/util.py", line 114, in typechecked_function
    return func(*args, **kwargs)
  File "/app/archivebox/extractors/media.py", line 75, in save_media
    index_texts = [
  File "/app/archivebox/extractors/media.py", line 76, in <listcomp>
    text_file.read_text(encoding='utf-8').strip()
  File "/usr/local/lib/python3.10/pathlib.py", line 1133, in read_text
    return f.read()
  File "/usr/local/lib/python3.10/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb3 in position 264619: invalid start byte

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

Issue #999, opened 2022-07-16
Bug: UnicodeDecodeError when archiving site


InnovativeInventor:
#### Describe the bug
Archiving a certain website (https://www.thedrive.com/the-war-zone/new-radars-are-giving-old-air-force-f-16s-capabilities-like-never-before) results in a `UnicodeDecodeError` error, which halts the update process. Ideally, there should be a way to skip errored archives.

#### Steps to reproduce
```
archivebox init
archivebox setup     # interactive setup
archivebox add www.thedrive.com/the-war-zone/new-radars-are-giving-old-air-force-f-16s-capabilities-like-never-before
```

#### Screenshots or log output

```
[+] [2022-07-16 03:26:32] "www.thedrive.com/the-war-zone/new-radars-are-giving-old-air-force-f-16s-capabilities-like-never-before"
    https://www.thedrive.com/the-war-zone/new-radars-are-giving-old-air-force-f-16s-capabilities-like-never-before
    > ./archive/1657940207.365172
      > title
      > favicon
      > headers
      > singlefile
        Extractor timed out after 60s.
        Run to see full output:
            [redacted]

      > pdf
      > screenshot
      > dom
      > wget
      > readability
      > mercury
      > media
    ! Failed to archive link: Exception: Exception in archive_methods.save_media(Link(url=https://www.thedrive.com/the-war-zone/new-radars-are-giving-old-air-force-f-16s-capabilities-like-never-before))

Traceback (most recent call last):
  File "/redacted/.pyenv/versions/3.9.10/lib/python3.9/site-packages/archivebox/extractors/__init__.py", line 109, in archive_link
    result = method_function(link=link, out_dir=out_dir)
  File "/redacted/.pyenv/versions/3.9.10/lib/python3.9/site-packages/archivebox/util.py", line 114, in typechecked_function
    return func(*args, **kwargs)
  File "/redacted/.pyenv/versions/3.9.10/lib/python3.9/site-packages/archivebox/extractors/media.py", line 74, in save_media
    index_texts = [
  File "/redacted/.pyenv/versions/3.9.10/lib/python3.9/site-packages/archivebox/extractors/media.py", line 75, in <listcomp>
    text_file.read_text(encoding='utf-8').strip()
  File "/redacted/.pyenv/versions/3.9.10/lib/python3.9/pathlib.py", line 1267, in read_text
    return f.read()
[...]

Issue #998, opened 2022-07-12
Question: YouTube videos not completing media archive method


jansendotsh:
Howdy 👋 I set up an ArchiveBox instance on a public facing VPS with docker-compose (using just ArchiveBox & Sonic). Whenever I send over a YouTube video, it seems to get wedged in a "not yet archived" state while the media archive method appears to not even kick off. I dug into the container's logs a bit and there looks to be an issue starting the `media.py` tool:

```
archivebox_1  | Internal Server Error: /add/
archivebox_1  | Traceback (most recent call last):
archivebox_1  |   File "/app/archivebox/extractors/__init__.py", line 109, in archive_link
archivebox_1  |     result = method_function(link=link, out_dir=out_dir)
archivebox_1  |   File "/app/archivebox/util.py", line 114, in typechecked_function
archivebox_1  |     return func(*args, **kwargs)
archivebox_1  |   File "/app/archivebox/extractors/media.py", line 74, in save_media
archivebox_1  |     index_texts = [
archivebox_1  |   File "/app/archivebox/extractors/media.py", line 75, in <listcomp>
archivebox_1  |     text_file.read_text(encoding='utf-8').strip()
archivebox_1  |   File "/usr/local/lib/python3.9/pathlib.py", line 1257, in read_text
archivebox_1  |     return f.read()
archivebox_1  |   File "/usr/local/lib/python3.9/codecs.py", line 322, in decode
archivebox_1  |     (result, consumed) = self._buffer_decode(data, self.errors, final)
archivebox_1  | UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd5 in position 414699: invalid continuation byte
archivebox_1  | The above exception was the direct cause of the following exception:
archivebox_1  | Traceback (most recent call last):
archivebox_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
archivebox_1  |     response = get_response(request)
archivebox_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
archivebox_1  |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
archivebox_1  |   File "/usr/local/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
archivebox_1  |     return self.dispatch(request, *args, **kwargs)
[...]

Issue #991, opened 2022-06-17
Bug: Docker install not able to save YouTube videos (media failure)


CorneliousJD:
#### Describe the bug
When using a basic Docker install it fails to save YouTube videos out of the box.

#### Steps to reproduce
Setup Docker install, archive any YouTube video, and it will fail to save the actual video files.

#### Screenshots or log output
Sometimes it will say it failed to retrieve media, other times it will say success but there's no actual video files saved.

#### ArchiveBox version
:latest

```logs
ArchiveBox v0.6.3
Cpython Linux Linux-5.15.46-Unraid-x86_64-with-glibc2.31 x86_64
IN_DOCKER=True DEBUG=False IS_TTY=True TZ=UTC SEARCH_BACKEND_ENGINE=ripgrep

[i] Dependency versions:
 √  ARCHIVEBOX_BINARY     v0.6.3          valid     /usr/local/bin/archivebox                                                   
 √  PYTHON_BINARY         v3.10.4         valid     /usr/local/bin/python3.10                                                   
 √  DJANGO_BINARY         v3.1.14         valid     /usr/local/lib/python3.10/site-packages/django/bin/django-admin.py          
 √  CURL_BINARY           v7.74.0         valid     /usr/bin/curl                                                               
 √  WGET_BINARY           v1.21           valid     /usr/bin/wget                                                               
 √  NODE_BINARY           v17.9.0         valid     /usr/bin/node                                                               
 √  SINGLEFILE_BINARY     v0.3.16         valid     /node/node_modules/single-file/cli/single-file                              
 √  READABILITY_BINARY    v0.0.2          valid     /node/node_modules/readability-extractor/readability-extractor              
 √  MERCURY_BINARY        v1.0.0          valid     /node/node_modules/@postlight/mercury-parser/cli.js                         
 √  GIT_BINARY            v2.30.2         valid     /usr/bin/git                                                                
 √  YOUTUBEDL_BINARY      v2022.04.08     valid     /usr/local/bin/yt-dlp                                                       
 √  CHROME_BINARY         v101.0.4951.41  valid     /usr/bin/chromium                                                           
 √  RIPGREP_BINARY        v12.1.1         valid     /usr/bin/rg                                                                 

[i] Source-code locations:
 √  PACKAGE_DIR           24 files        valid     /app/archivebox                                                             
 √  TEMPLATES_DIR         4 files         valid     /app/archivebox/templates                                                   
[...]

PR #1026, opened 2022-09-12
Warn on broken steps, use yt-dlp to avoid youtube-dl errors, and don't crash on bad UTF-8

turian:
# Summary

* If some step is broken, we issue a warning instead of crashing.
* We use yt-dlp instead of youtube-dl by default.
* On bad unicode, we don't crash.

# Quickest workaround for many people, until this is merged

Add this to ArchiveBox.conf:
```
YOUTUBEDL_BINARY=/usr/bin/yt-dlp
```

If that doesn't work, you can use my Docker `turian/archivebox:kludge-984-UTF8-bug` which includes this patch, instead of `archivebox/archivebox` for now.

@jgoerzen I think you also said this bug was a showstopper for you

# Try it out

This finally works with this patch:

`archivebox add 'https://www.ashra.com/news.php?m=A'`

# Related issues

Should close these issues:
* UnicodeDecodeError: https://github.com/ArchiveBox/ArchiveBox/issues/1014 and https://github.com/ArchiveBox/ArchiveBox/issues/984 https://github.com/ArchiveBox/ArchiveBox/issues/999
* Not able to save YouTube videos which might be UnicodeDecodeError but are probably also `youtube-dl` errors: https://github.com/ArchiveBox/ArchiveBox/issues/991 https://github.com/ArchiveBox/ArchiveBox/issues/998 

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

# `youtube-dl` -> `yt-dlp`

So I began by changing the `archive_link` from a crashing exception to a warning. This gave me better diagnostics than the exceptions you see in the issues above. I observed the following:

```
       Extractor failed:
             Failed to save media
        Run to see full output:
            cd /home/archivebox/archivebox-turian/ArchiveBox/pipdata/archive/1663011964.432896;
            youtube-dl --write-description --write-info-json --write-annotations --write-thumbnail --no-call-home --write-sub --all-subs --write-auto-sub --convert-subs=srt --yes-playlist --continue --ignore-errors --no-abort-on-error --geo-bypass --add-metadata --max-filesize=750m https://www.ashra.com/news.php?m=A
```

Inspecting that, I saw:

```
$ youtube-dl --write-description --write-info-json --write-annotations --write-thumbnail --no-call-home --write-sub --all-subs --write-auto-sub --convert-subs=srt --yes-playlist --continue --ignore-errors --no-abort-on-error --geo-bypass --add-metadata --max-filesize=750m https://www.ashra.com/news.php?m=A
Usage: youtube-dl [OPTIONS] URL [URL...]

youtube-dl: error: no such option: --no-abort-on-error
```

Oups. `youtube-dl` doesn't have `--no-abort-on-error`. I think it only has `--abort-on-error` but I'm not entirely sure what the default behavior is.

I switched [youtube-dl](https://github.com/ytdl-org/youtube-dl) to [yt-dlp](https://github.com/yt-dlp/yt-dlp) as the default, which is a more actively maintained fork that pulls upstream from youtube-dl constantly. It also has option `--no-abort-on-error` so now the media download works.

It operates essentionally identically (and FASTER) to youtube-dl.

The above youtube-dl options are present, with the following caveats:
* We aren't doing this, but "When --embed-subs and --write-subs are used together, the subtitles are written to disk and also embedded in the media file. You can use just --embed-subs to embed the subs and automatically delete the separate file."
* We aren't doing this, but "--add-metadata attaches the infojson to mkv files in addition to writing the metadata when used with --write-info-json. Use --no-embed-info-json or --compat-options no-attach-info-json to revert this"
* `--write-annotations`: "No supported site has annotations now"

I haven't pushed yt-dlp default changes to the submodules yet:
`deb_dist, brew_dist, docker, docs, etc/ArchiveBox.conf.default, pip_dist`
but I'd like to. I would even go so far as to deprecate `youtube-dl`, but perhaps that's too radical for some people.

# UnicodeDecodeError fixes

A common complaint. In `media.py`, we used to have:
```
text_file.read_text(encoding='utf-8').strip()
```
This is strict and errors cause a crash. I have changed the behavior to [xmlcharrefreplace](https://github.com/ArchiveBox/ArchiveBox/pull/1026/files#diff-9025388b81475556336588c7ee9ba7dbee7e2afb73c962c1fce3d81bc6ca9dd8R77) ("Characters not supported by the encoding are replaced with the appropriate XML character reference &#nnn;.") which seemed the best to me, but you can read about other options in my comment there or in more detail at Python's [open](https://docs.python.org/3/library/functions.html#open) function documentation.

My preferred workaround, because sometimes things say they are utf-8 but are actually a different encoding, would be this:
1) Try utf-8 in strict mode. If clean, then good.
2) If error, use [chardet](https://chardet.readthedocs.io/en/latest/usage.html#basic-usage) to guess the encoding. (It's a little slow, so it's better as a fallback option.) Attempt to decode to Unicode using the guessed encoding in strict mode. If clean, then good.
3) Worst case: Either use utf-8 or the highest confidence `chardet` guessed encoding, and decode it using `xmlcharrefreplace` for errors.

This would be a separate PR, if you approve of me including `chardet` as a pip dependency here and in all submodules (same as with yt-dlp).

# Postscript

I ran `flake8` and caught all the flakes in the code I introduced, as far as I know. Since I am a new committer, CI/CD won't run for me here.


archivebox/extractors/__init__.py
```diff
@@ -126,11 +128,28 @@ def archive_link(link: Link, overwrite: bool=False, methods: Optional[Iterable[s
...
+                with open(ERROR_LOG, "a", encoding='utf-8') as f:
+                    command = ' '.join(sys.argv)
+                    ts = datetime.now(timezone.utc).strftime('%Y-%m-%d__%H:%M:%S')
+                    f.write(("\n" + 'Exception in archive_methods.save_{}(Link(url={})) command={}; ts={}'.format(
+                        method_name,
+                        link.url,
+                        command,
+                        ts
+                    ) + "\n"))
+                    #f.write(f"\n> {command}; ts={ts} version={config['VERSION']} docker={config['IN_DOCKER']} is_tty={config['IS_TTY']}\n")
```
turian, line 152: Should there be a `break` here to exit the loop?

notevenaperson:
+1 for yt-dlp

setup.py
```diff
@@ -42,6 +42,7 @@
     "django-extensions>=3.0.3",
     "dateparser>=1.0.0",
     "youtube-dl>=2021.04.17",
+    "yt-dlp>=2021.4.11",
```
turian, line 45: I arbitrarily picked this version number to be the first datestamp ahead of the youtube-dl version.

turian:
A few more notes:
* `yt-dlp` is so much faster than `youtube-dl` that you get throttled if you download all the autogenerated subtitles. So I simply disabled them, and only generate "real" subtitles and captions.

archivebox/config.py
```diff
@@ -144,12 +144,19 @@
...
                                                                 '--convert-subs=srt',
                                                                 '--yes-playlist',
                                                                 '--continue',
-                                                                '--ignore-errors',
+                                                                # This flag doesn't exist in youtube-dl
+                                                                # only in yt-dlp
                                                                 '--no-abort-on-error',
+                                                                # --ignore-errors must come AFTER
+                                                                # --no-abort-on-error
+                                                                # https://github.com/yt-dlp/yt-dlp/issues/4914
```
turian, line 158: The order of these flags is---confusingly---quite important.

https://github.com/yt-dlp/yt-dlp/issues/4914

WakeReality:
I'm using docker-compose to run ArchiveBox. How do I get yt-dlp inside the docker container?

pirate:
Yt-dlp is already the default in the new beta 0.4.3 `latest` builds.

setup.py
pirate: you can bump this to the latest version whenever you touch these files (as long as it works), no need to match older versions of chrome/youtubledl/ffmpeg/git/wget, only django and a couple other edge cases are the ones that cant be bumped

pirate (approved):

turian:
boom!

PR #1059, opened 2022-12-05
Fix some encoding-related issue of readability

mattholy:
# Summary

This PR fixes some issue of readability by added `<meta charset="UTF-8">` to the top of `content.html`
image

Tested on Microsoft Edge and Google Chrome
# Changes these areas

- [x] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/extractors/readability.py
+++ b/archivebox/extractors/readability.py
@@ -77,7 +77,7 @@ def save_readability(link: Link, out_dir: Optional[str]=None, timeout: int=TIMEO
 
         output_folder.mkdir(exist_ok=True)
         readability_content = result_json.pop("textContent") 
-        atomic_write(str(output_folder / "content.html"), result_json.pop("content"))
+        atomic_write(str(output_folder / "content.html"), '<meta charset="UTF-8">\n' + result_json.pop("content"))
         atomic_write(str(output_folder / "content.txt"), readability_content)
         atomic_write(str(output_folder / "article.json"), result_json)
 


notevenaperson:
I have this issue as well, on many pages. [Beautifulracket.com](https://beautifulracket.com):
image

However, it looks fine in the index, because the index page has `<meta charset="utf-8"` declaration.
image

Perhaps surprisingly, context.txt is broken as well, so this pull request alone won't fix that.
image

## How to reproduce: 

``` sh
docker exec archivebox-archivebox-1 wget 'https://beautifulracket.com/' -O beautifulracket.html
docker exec archivebox-archivebox-1 readability-extractor beautifulracket.html 'https://beautifulracket.com/' > some_article.json
```

You can see some characters don't render.
image

## Leads
a) https://readability.readthedocs.io/en/latest/api.html#readability.encoding.fix_charset
b) https://github.com/dankito/Readability4J#output-encoding "Readability.js returns its output in a `<div>` element, and the only way to set the encoding in HTML is in a `<head><meta charset="">` tag."

pirate:
I'm not super comfortable modifying the output html like this, is there a way to do this by sending a header through django instead of modifying the archived files?

notevenaperson:
> I'm not super comfortable modifying the output html like this, is there a way to do this by sending a header through django instead of modifying the archived files?

Yes. Some sites only send a header, as [I discovered](https://github.com/ArchiveBox/readability-extractor/pull/5/commits/f9c463e41e21678305392868e69c807056a661be) when I dug into this. The header is "content-type: text/html;charset=utf-8".

pirate:
Great, we can add that as a middleware to Django that will serve this header with archived files when appropriate. That way we don't have to modify the on-disk archived data.

notevenaperson:
> Great, we can add that as a middleware to Django that will serve this header with archived files when appropriate. That way we don't have to modify the on-disk archived data.

But how do you determine when it's appropriate? Not all html pages are utf-8. The page stripped by Readability no longer has all the information necessary to know for sure what the character set the page intended to be in. The character set information has to be determined *before* or during the readability-extractor step.

notevenaperson:
> > Great, we can add that as a middleware to Django that will serve this header with archived files when appropriate. That way we don't have to modify the on-disk archived data.
> 
> But how do you determine when it's appropriate? Not all html pages are utf-8. The page stripped by Readability no longer has all the information necessary to know for sure what the character set the page intended to be in. The character set information has to be determined _before_ or during the readability-extractor step.

I had an idea, the readability-extractor json output could have a "content-charset" key, that later Django could choose to read, to send the correct header. That way we don't alter the html.
```json
{
  "title": "",
  "byline": null,
  "dir": null,
  "content": "<div id=\"readability-page-1\" class=\"page\">©\n</div>",
  "content-charset": "UTF-16LE"
  "textContent": "©\n",
  "length": 2,
  "siteName": null
}
```

pirate:
I'm going to close this for now as the issue remains unsolved until we have better charset detection upstream from readability.

If anyone wants to take a stab at fixing this, it will involve a PR to readability first to get them to add the detected charset in their JSON output, then we can revisit using it to add some headers for ArchiveBox viewers.

Issue #179, opened 2019-03-19
Add official support for taking multiple snapshots of websites over time


pirate:
This is by far the most requested feature.

People want an easy way to take multiple snapshots of websites over time.

> Here's how archive.org does it
> <img width="500" alt="Screenshot 2024-03-09 at 18 58 25" src="https://github.com/ArchiveBox/ArchiveBox/assets/4463796/92a5cdf9-9fd1-4fb9-9903-0c28c47e7884">

---

For people finding this issue via Google / incoming links, if you want a hacky solution to take a second snapshot of a site, you can add the link with a new hash and it will be treated as a new page and a new snapshot will be taken:
```bash
echo https://example.com/some/page.html#archivedate=2019-03-18 | archivebox add
# then to re-shapshot it on another day...
echo https://example.com/some/page.html#archivedate=2019-03-22 | archivebox add
```

*Edit: as of v0.6 there is now a button in the UI to do this ^*  
![Screen Shot 2021-04-09 at 8 51 35 p](https://user-images.githubusercontent.com/511499/114264320-1f1da200-99b8-11eb-986f-bd8c808e5110.png)

n0ncetonic:
Looking forward to this feature. Thanks for the hacky workaround as well, I have a few pages I'd like to continue monitoring for new content but I was worried about the implications of my current backup being overwritten by a 404 page if the content went down.

pirate:
I just updated the [README](https://github.com/pirate/ArchiveBox#Overview) to make the current behavior clearer as well:

> Running `./archive` adds only new, unique links into your data folder on each run. Because it will ignore duplicates and only archive each link the first time you add it, you can schedule it to [run on a timer](https://github.com/pirate/ArchiveBox/wiki/Scheduled-Archiving) and re-import all your feeds multiple times a day. It will run quickly even if the feeds are large, because it's only archiving the newest links since the last run.  For each link, it runs through all the archive methods. Methods that fail will save `None` and be automatically retried on the next run, methods that succeed save their output into the data folder and are never retried/overwritten by subsequent runs.  Support for saving multiple snapshots of each site over time will be [added soon](https://github.com/pirate/ArchiveBox/issues/179) (along with the ability to view diffs of the changes between runs).

alex9099:
Any updates on this? It would be really nice if it was possible to have versions, like the waybackmachine does :)

pirate:
You can accomplish this right now by adding a hash at the end of the URL, e.g.

[...]

PR #1110, opened 2023-03-01
Add resnapshot option to `add` and `schedule` functions

kelvinburke:
# Summary

This PR adds a `--resnapshot` flag to the `add` and `schedule` functions so that one can re-snapshot from the CLI.
 - Added `add_timestamp_to_links` to `archivebox.index.__init__` which timestamps the url in the same way as `resnapshot_snapshot`

# Related issues

As discussed in https://github.com/ArchiveBox/ArchiveBox/issues/179

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [x] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/cli/archivebox_add.py
+++ b/archivebox/cli/archivebox_add.py
@@ -75,6 +75,12 @@ def main(args: Optional[List[str]]=None, stdin: Optional[IO]=None, pwd: Optional
         help="Re-archive URLs from scratch, overwriting any existing files"
     )
     parser.add_argument(
+        "--resnapshot",
+        default=False,
+        action="store_true",
+        help="Re-archive URLs from scratch, creating a new snapshot timestamped with the current time"
+    )
+    parser.add_argument(
         "--init", #'-i',
         action='store_true',
         help="Init/upgrade the curent data directory before adding",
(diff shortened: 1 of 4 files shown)


kelvinburke:
I am thinking it could be useful for `update` to also have the `--resnapshot` flag. Then `update --resnapshot` would re-snapshot the entire library.

kelvinburke:
One more thing that `add_timestamp_to_links` is do is deduplicate the links, as multiple of the same link with different timestamps could get passed in, then we would get multiple with the same new timestamp.

pirate:
This is great, thanks! Can you make the Admin UI Resnapshot button re-use the same `add_timestamp_to_links` logic?

pirate:
Going to close this for now as stale as I need to clean out old PRs in order to prepare a new release. Thanks for your work on it so far and I welcome you to open a new PR if you still want to add this feature.

Any future attempt to implement this should centralize the resnapshot logic between the admin UI button and the CLI flags. It should also not append `#<timestamp>` hashes to URLs that are seen for the first time. (i.e. it should only "re-"snapshot URLs that have been previously snapshotted, if a URL is new it doesn't need the timestamp appended)

PR #1115, opened 2023-03-08
add systemd config

fa0311:
# Summary

Add an example configuration file.
This configuration file is what I actually use on UbuntuServer.

# Related issues

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/etc/archive-box.service
+++ b/etc/archive-box.service
@@ -1,0 +1,28 @@
+#
+# These are just a few examples of settings
+#
+# Items that probably need to be rewritten depending on the environment:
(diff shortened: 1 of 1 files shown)


pirate:
looks good, thanks!

do you mind swapping all the `archive-box` strings with `archivebox`? we don't use the dash form anywhere else so it makes sense to be consistent

fa0311:
of course

PR #1120, opened 2023-03-14
add CHROME_TIMEOUT args

fa0311:
# Summary

- add `CHROME_TIMEOUT` args

# Environment

`Chromium 111.0.5563.64 snap`

# Why

Chromium `--timeout` argument doesn't actually do timeout behavior.
For example, when `--screenshot` argument is given, it delays the shooting for the time specified by the timeout.
Not sure if this behavior is caused by chromium v111.

# Note

CHROME_TIMEOUT is set to 0 by default.
For example, when `--screenshot` argument is given, If timeout is set to 0, the image is captured at the same time DOMContentLoaded fires, so some Web sites may not be captured as intended.
When set to 1 or higher, the possibility of getting the desired photo increases dramatically.

# Related

https://github.com/ArchiveBox/ArchiveBox/commit/606fa397a42e2821d8879b6274e056c8f75ccbe0

# Changes these areas

- [x] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/config.py
+++ b/archivebox/config.py
@@ -139,6 +139,7 @@ CONFIG_SCHEMA: Dict[str, ConfigDefaultDict] = {
         'COOKIES_FILE':             {'type': str,   'default': None},
         'CHROME_USER_DATA_DIR':     {'type': str,   'default': None},
 
+        'CHROME_TIMEOUT':           {'type': int,  'default': 0},
         'CHROME_HEADLESS':          {'type': bool,  'default': True},
         'CHROME_SANDBOX':           {'type': bool,  'default': lambda c: not c['IN_DOCKER']},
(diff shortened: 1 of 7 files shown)


pirate:
Thanks!

I confirmed it seems to be a recent change they made: https://developer.chrome.com/articles/new-headless/

fa0311:
> I confirmed it seems to be a recent change they made: https://developer.chrome.com/articles/new-headless/

This seems to be implemented in v112, so it doesn't seem to affect v111, but v111 seems to be unstable due to this.
For example, --window-size does not work,

fa0311:
I have set up Issues regarding this. #1122

Issue #1117, opened 2023-03-09
Feature request: Add ability to point ArchiveBox to a custom favicon service instead of Google


pirate:
Just found out that https://github.com/mat/besticon exists.  This would allow us to point ArchiveBox at a self-hosted replacement for the Google favicon service we currently use.

The current solution has always mildly irked me as a suboptimal way we leak archived domains to a 3rd party, which is why I document how to disable it here: https://docs.archivebox.io/en/latest/Security-Overview.html#stealth-mode, but it would be nice to provide an alternative other than disabling it entirely.

I'm imagining something like:

`archivebox config --set FAVICON_PROVIDER='https://www.google.com/s2/favicons?domain={}'` (the default)

that can be set to anything, e.g.:

`archivebox config --set FAVICON_PROVIDER='https://someselfhostedservice.example.com/?q={}&format=png&size=64x64'`

pirate:
closing because PR was merged, thanks!

PR #1146, opened 2023-05-06
Add FAVICON_PROVIDER option for custom favicon service

mrled:
# Summary

Add `FAVICON_PROVIDER` option to allow configuring any favicon service.

Tested with [The Favicon Finder](https://besticon-demo.herokuapp.com/) (a demo deployment of [mat/besticon](https://github.com/mat/besticon) mentioned in the related issue), and also with [Favicon Kit](faviconkit.com).

e.g.

```sh
archivebox config --set FAVICON_PROVIDER='https://besticon-demo.herokuapp.com/icon?url={}&size=32..64..64'
```

# Questions

I noticed that the Google URL was hard-coded in two places: `archivebox/extractors/favicon.py` and `archivebox/index/schema.py`. I made changes in both, but I'm not sure if I should have done the latter -- the top of that file says not to add any more features to the file. Is it better to leave it as is, with the old hard-coded Google URL, or change it?

# Related issues

#1117

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/config.py
+++ b/archivebox/config.py
@@ -183,7 +183,8 @@ CONFIG_SCHEMA: Dict[str, ConfigDefaultDict] = {
                                                                 '--compressed'
                                                                ]},
         'GIT_ARGS':                 {'type': list,  'default': ['--recursive']},
-        'SINGLEFILE_ARGS':          {'type': list,  'default' : None}
+        'SINGLEFILE_ARGS':          {'type': list,  'default' : None},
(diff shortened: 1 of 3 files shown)


pirate:
This is perfect, thanks! the change to `schema.py` is ok, I think the new behavior is expected in this case.

PR #1197, opened 2023-07-31
bin_version means to modify, not replace environ

overhacked:
# Summary

The `bin_version` function means to modify the environment, not replace it entirely. Fixes bugs that occur when it wipes out the PATH environment variable, such as when running in a virtual environment. E.g. results in bin_version not being able to invoke `node` when it exists on the `PATH`.

# Related issues

*None*

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/config.py
+++ b/archivebox/config.py
@@ -690,7 +690,8 @@ def bin_version(binary: Optional[str]) -> Optional[str]:
         return None
 
     try:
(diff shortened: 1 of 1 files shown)


pirate:
Wow nice catch, this was bugging me for a long time but I never found the source of the issue!

Thanks!

PR #1201, opened 2023-08-04
Remove/detect empty media extractor directory

overhacked:
If the media extraction fails, try to remove an empty media directory,
ignoring any errors. When retrying extractors (e.g. `--update`), detect
an empty media directory and run the extractor; recovers from failed
media extractions prior to this commit.

### DIFF
--- a/archivebox/extractors/media.py
+++ b/archivebox/extractors/media.py
@@ -26,7 +26,9 @@ def should_save_media(link: Link, out_dir: Optional[Path]=None, overwrite: Optio
         return False
 
     out_dir = out_dir or Path(link.link_dir)
-    if not overwrite and (out_dir / 'media').exists():
+    media_dir = out_dir / 'media'
+    media_dir_has_files = media_dir.exists() and any(media_dir.iterdir())
+    if not overwrite and media_dir_has_files:
         return False
 
     return SAVE_MEDIA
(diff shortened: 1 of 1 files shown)


archivebox/extractors/media.py
```diff
@@ -26,7 +26,9 @@ def should_save_media(link: Link, out_dir: Optional[Path]=None, overwrite: Optio
         return False
 
     out_dir = out_dir or Path(link.link_dir)
-    if not overwrite and (out_dir / 'media').exists():
+    media_dir = out_dir / 'media'
+    media_dir_has_files = media_dir.exists() and any(media_dir.iterdir())
```
pirate, line 30:
```suggestion
    media_dir_has_files = media_dir.exists() and any(filename for filename in media_dir.iterdir() if not filename.startswith('.'))
```
Should we exclude `.DS_Store` files and other hidden temp files here?

pirate:
after further review I realize the original design was intentional. The empty directory was being used as a record of the "tried to extract media but none was found" state. I don't necessarily want to retry youtube-dl on every URL that produces no media, as many URLs legitimately contain no media and don't need to be retried potentially hundreds of times (every time `archivebox update` is run in the future).

Is there some way we can differentiate a transient failure that might be fixed later vs a hard fail like "no media could possibly be found on this URL and it's highly unlikely that any might be found in the future".

Issue #874, opened 2021-10-05
Bug: URL getting truncated at underscore or parens


zblesk:
#### Describe the bug

After pasting a URL, it gets truncated at `_` or `(` and the wrong page is downloaded.

#### Steps to reproduce

I enter a URL, like this one: `https://vocaloidlyrics.fandom.com/wiki/%E4%BA%BA%E6%9F%B1%E3%82%A2%E3%83%AA%E3%82%B9_(Hitobashira_Alice)`
by going to the Web UI, clicking ADD, pasting this URL on a single line and confirming.

The archival process runs; however, it seems some of the archival methods truncate the URL to `https://vocaloidlyrics.fandom.com/wiki/%E4%BA%BA%E6%9F%B1%E3%82%A2%E3%83%AA%E3%82%B9_`
and others to `https://vocaloidlyrics.fandom.com/wiki/%E4%BA%BA%E6%9F%B1%E3%82%A2%E3%83%AA%E3%82%B9`

The "original URL" in the UI ends up being `https://vocaloidlyrics.fandom.com/wiki/%E4%BA%BA%E6%9F%B1%E3%82%A2%E3%83%AA%E3%82%B9_` 

#### Screenshots or log output

```
[...]

Issue #864, opened 2021-09-28
Bug: URLs with parentheses are mishandled by `archivebox add`


WesleyAC:
#### Describe the bug

URLs like `https://en.wikipedia.org/wiki/APL_(programming_language)` are silently converted to `https://en.wikipedia.org/wiki/APL_`.

#### Steps to reproduce

1. `archivebox init`
2. `echo "https://en.wikipedia.org/wiki/APL_(programming_language)" | archivebox add` OR `archivebox add "https://en.wikipedia.org/wiki/APL_(programming_language)"`
3. `archivebox list` now shows `https://en.wikipedia.org/wiki/APL_ "APL - Wikipedia"`, which is incorrect.

#### ArchiveBox version

```logs
ArchiveBox v0.6.2
Cpython Linux Linux-5.10.66-x86_64-with-glibc2.33 x86_64
IN_DOCKER=False DEBUG=False IS_TTY=True TZ=UTC SEARCH_BACKEND_ENGINE=ripgrep

[i] Dependency versions:
 √  ARCHIVEBOX_BINARY     v0.6.2          valid     /nix/store/xfqwr74qli0fm186dj5006960mn5pw38-archivebox-0.6.2/bin/archivebox 
 √  PYTHON_BINARY         v3.9.6          valid     /nix/store/i1m8r7mv8h47wr850cdsxksy22lv6gsz-python3-3.9.6/bin/python3.9     
 √  DJANGO_BINARY         v3.1.7          valid     /nix/store/889dn8m5c562vy949frmxdna4kxfm8rf-python3.9-Django-3.1.7/lib/python3.9/site-packages/django/bin/django-admin.py
 √  CURL_BINARY           v7.76.1         valid     /nix/store/blaac45yvgljsy15jdxgvxxqs6w5yhqj-curl-7.76.1-bin/bin/curl        
 -  WGET_BINARY           -               disabled  /run/current-system/sw/bin/wget                                             
 √  NODE_BINARY           v14.17.6        valid     /nix/store/zqgmd79n5p0mdaw4sbvkv7gvrmks76a2-nodejs-14.17.6/bin/node         
 √  SINGLEFILE_BINARY     v0.3.31         valid     ./node_modules/single-file/cli/single-file                                  
 √  READABILITY_BINARY    v0.0.3          valid     ./node_modules/readability-extractor/readability-extractor                  
[...]

Issue #287, opened 2019-10-18
Incorrect handling of parentheses in url


Plausible0Bug:
#### Describe the bug

Attempt to archive `https://en.m.wikipedia.org/wiki/A_good_day_to_die_(phrase)` will archive `https://en.m.wikipedia.org/wiki/A_good_day_to_die_` instead.

#### Steps to reproduce

Run `echo 'https://en.m.wikipedia.org/wiki/A_good_day_to_die_(phrase)' | /bin/archive'

#### Screenshots or log output

```
echo "https://en.m.wikipedia.org/wiki/A_good_day_to_die_(phrase)" | sudo docker run -i --rm -v ~/ArchiveBox:/data:rw,z nikisweeting/archivebox
[*] [2019-10-18 03:09:37] Parsing new links from output/sources/stdin-1571368177.txt...
    > Adding 1 new links to index (parsed import as Plain Text)
[*] [2019-10-18 03:09:37] Saving main index files...
    √ /data/index.json
    √ /data/index.html
[▶] [2019-10-18 03:09:37] Updating content for 1 pages in archive...

[+] [2019-10-18 03:09:37] "https://en.m.wikipedia.org/wiki/A_good_day_to_die_"
    https://en.m.wikipedia.org/wiki/A_good_day_to_die_
    > /data/archive/1571368177
      > title
      > favicon
      > wget
      > pdf
      > screenshot
      > dom
      > media
      > archive_org
[√] [2019-10-18 03:09:59] Update of 1 pages complete (22.14 sec)
    - 0 links skipped
    - 1 links updated
    - 0 links had errors
    To view your archive, open: /data/index.html
[*] [2019-10-18 03:09:59] Saving main index files...
    √ /data/index.json
    √ /data/index.html
```

#### Software versions

 - OS:                        ([e.g. macOS 10.14] the operating system you're running ArchiveBox on)
 - ArchiveBox version:        (`git rev-parse HEAD | head -c7` [e.g. d798117] commit ID of the version you're running)
 - Python version:            (`python3 --version` [e.g. 3.7.0])
 - Chrome version:            (`chromium-browser --version` [e.g. 73.1.2.3] if relevant to bug)

pirate:
Thanks for reporting! We're aware of this issue and it's trickier than it looks on the surface, for now you can import your links as JSON, XML, RSS, etc to avoid this problem (let me know if you need help with doing that). 

Duplicate: https://github.com/pirate/ArchiveBox/issues/235

Issue #235, opened 2019-05-06
Parsing: Sources that aren't URL-encoded can contain URLs with brackets, parens, and other separator symbols


anarcat:
#### Describe the bug

I was trying to archive this Wikipedia page:

https://en.wikipedia.org/wiki/Insert_(effects_processing)

Archivebox sees it as:

https://en.wikipedia.org/wiki/Insert_

Bad box! No cookie! :)

#### Steps to reproduce

echo 'https://en.wikipedia.org/wiki/Insert_(effects_processing)' | archivebox add

#### Screenshots or log output

```
[+] [2019-05-06 22:21:48] "en.wikipedia.org/wiki/Insert_"
    https://en.wikipedia.org/wiki/Insert_
    > ./archive/1557178364.70
      > title
      > favicon
      > wget
      > pdf
      > screenshot
      > dom
      > media
      > archive_org
```

#### Software versions

 - OS: Debian buster 10
 - ArchiveBox version: 0.4.1 installed from pypi
 - Python version: 3.7.3 something
 - Chrome version: irrelevant?

#### Workaround

Run `sed 's/(/%28/;s/)/%29/'` over the list of URLs, although I suspect more special characters might be affected.

pirate:
Ah damn, I explicitly stop parsing at those characters because of cases like this:
```python
>>>parse('This is some text in a markdown file with [a link](https://example.com)/[some other link](https://example.com/b/)...')
['https://example.com)/[some', 'https://example.com/b/)...']
```

The current regex looks like this:
```python
URL_REGEX = re.compile(
    r'http[s]?://'                    # start matching from allowed schemes
    r'(?:[a-zA-Z]|[0-9]'              # followed by allowed alphanum characters
    r'|[$-_@.&+]|[!*\(\),]'           #    or allowed symbols
    r'|(?:%[0-9a-fA-F][0-9a-fA-F]))'  #    or allowed unicode bytes
    r'[^\]\[\(\)<>\""\'\s]+',         # stop parsing at these symbols
    re.IGNORECASE,
)
```

I don't see any easy solutions, but there are a few crappy options:
 - require all URLs be URL-encoded before parsing
 - add a config option to enable parsing of `()` and `[]` in URLs
 - attempt to autodetect the import format and parse accordingly (this is extremely hard)

[...]

PR #1205, opened 2023-08-08
Fix hyphen placement in util.URL_REGEX

overhacked:
# Summary

Incorrect hyphen placement in `URL_REGEX` was allowing it to match more characters than intended. In a regex character class, a literal hyphen can only appear as the first character in the class, or it will be interpreted as the delimiter of a range of characters.

The issue fixed here caused the range of characters from `[$-_]` be treated as valid URL characters, instead of the intended set of three characters `[-_$]`. The incorrect range interpretation inadvertantly included most ASCII punctuation, most importantly the angle brackets, square brackets, and single quote that the expression uses to mark the end of a match.

This causes the expression to match a URL that has a "hostname" portion beginning with one of the intended "stop parsing" characters. For example:

```
https://<b>www</b>.example.com/  # MATCHES but should not
https://[for example]            # MATCHES but should not
scheme='https://'                # MATCHES, including final quote, but should not
```

Some test cases have been added to the `URL_REGEX` assert in archivebox.parsers to cover this possibility.

# Related issues

There are other `URL_REGEX` issues (#235, #287, #864, #874), but none that this change directly impacts.

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
Thanks, good catch! My regex testing app didn't highlight this issue because it must've assumed it was intentional, so it slipped by me!

PR #1207, opened 2023-08-08
Add config for using timestamp as Snapshot.added

### LINKED ISSUE #785: Feature Request: Import bookmarks from Pinboard and keep their original timestamps?
## Type 

 - [ ] General question or discussion
 - [ ] Propose a brand new feature
 - [x] Request modification of existing behavior or design

## What is the problem that your feature request solves
I'd like to migrate from Pinboard to ArchiveBox. I have over 10,000 bookmarks since 2009. Export them as JSON and import them into ArchiveBox but found that all bookmarks have the same timestamps.

```bash
docker-compose run archivebox add < ~/archivebox/pinboard_export.2021.07.07_06.33.json
```

## Describe the ideal specific solution you'd want, and whether it fits into any broader scope of changes
The import process could keep the original timestamp that's already in the JSON export:

```json
...
{
  "href":"https:\/\/github.com\/yuliskov\/SmartTubeNext",
  "time":"2021-06-05T18:55:20Z",
  "shared":"no",
  "toread":"no",
  "tags":"youtube android"
},
```

## What hacks or alternative solutions have you tried to solve the problem?
Manually edit 10,000 imported bookmarks? I don't think so...

## How badly do you want this new feature?

 - [x] It's an urgent deal-breaker, I can't live without it
 - [ ] It's important to add it in the near-mid term future
 - [ ] It would be nice to have eventually

---

 - [ ] I'm willing to contribute [dev time](https://github.com/ArchiveBox/ArchiveBox#archivebox-development) / [money](https://github.com/sponsors/pirate) to fix this issue
 - [ ] I like ArchiveBox so far / would recommend it to a friend
 - [x] I've had a lot of difficulty getting ArchiveBox set up

overhacked:
# Summary

:boom: **IMPORTANT: This PR includes a Django migration.** :boom:

Adds ability to set `added` field to the imported `timestamp`. Useful when importing from RSS or other rich formats like JSON that have past timestamps in the import metadata, so the ArchiveBox interface can display the source imported times instead of a large batch all with the same added date.

## Open questions before merging
- **Should this behavior be default turned off?**
- Should it be configurable at all?
- Should it be turned on, but only when importing from a metadata-rich parser?

# Related issues

Fixes #785.

# Changes these areas

- [x] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
We already have a field for this, it's `Snapshot.bookmarked`!  `.added` is for when Django first became aware of the URL, and `.bookmarked` is for any timestamp read in from an import source like bookmarks, RSS, pocket feed, etc.  We need both to remain separate, but we could add an option to prefer displaying `.bookmarked` or `.added` in the UI in different places.

overhacked:
Ah ha, okay, I misunderstood the problem here! I'll rework to make a UI setting instead of an import setting.

pirate:
I'm going to close this for now because if it's implemented it's going to be quite different from this first approach.

If you keep working on this with the new approach do you mind opening it as a separate PR? Thanks!

Issue #554, opened 2020-11-28
Less "open" public website


shepner:
## Type 

 - [ ] General question or discussion
 - [ ] Propose a brand new feature
 - [x] Request modification of existing behavior or design

## What is the problem that your feature request solves

Occasionally I would like to share links to pages I have clipped.  However, the public ArchiveBox site is rather "open" (very easy for anyone to look at everything stored in there) and Id rather if that was not the case.  Im not asking for iron-clad security here but Id like to keep people (and bots) from snooping.

## Describe the ideal specific solution you'd want, and whether it fits into any broader scope of changes

Ideally: Implement granular access controls to the website which are tied to user accounts which can exist in an external system (ie Okta, Authelia, LDAP, etc).
Minimally: Make it so the directory structure is not publicly visible, provide an option to disable public access to the archive root, and an option to entirely disable public archive access.

## What hacks or alternative solutions have you tried to solve the problem?

Since I want to make ArchiveBox Internet accessible so I can archive sites from my mobile devices, the minimal solution above should provide a modicom of privacy and is better than expecting the user to somehow figure out how to do this via Nginx or the like.

## How badly do you want this new feature?

 - [ ] It's an urgent deal-breaker, I can't live without it
 - [x] It's important to add it in the near-mid term future
 - [ ] It would be nice to have eventually

---

 - [ ] I'm willing to contribute dev time / money to fix this issue
 - [x] I like ArchiveBox so far / would recommend it to a friend
 - [ ] I've had a lot of difficulty getting ArchiveBox set up

pirate:
Config options to manage which parts are public already exist:

![image](https://user-images.githubusercontent.com/511499/100522069-17323380-3176-11eb-9a06-7b3b7e3a42c0.png)
https://github.com/ArchiveBox/ArchiveBox/blob/e4d2ac432db2627cec7cfa5fb02e8b83bb4269bb/archivebox/config.py#L71

```bash
archivebox config --set PUBLIC_SNAPSHOTS=True
archivebox config --set PUBLIC_INDEX=False
archivebox config --set PUBLIC_ADD_VIEW=False
```

You can also manage the permissions on a per-user basis under http://127.0.0.1:8000/admin/core/user/

shepner:
Ok, I think I figured out what those settings do.  Are they documented somewhere?  Similarly, is there some documentation for the various user account permissions?  I havnt found them in the wiki.

pirate:
[...]

PR #1214, opened 2023-08-18
Add LDAP support

DanielBatteryStapler:
# Summary

This project uses Django, which has an LDAP authentication backend that can be enabled (if you install it). These changes add the required configuration options to select and configure this backend to be used instead of Django's internal user backend.

Example environment variable configuration:
```
LDAP: "True"
LDAP_SERVER_URI: "ldap://daniel-authenticator:3389"
LDAP_BIND_DN: "ou=archivebox,ou=services,dc=daniel-authenticator"
LDAP_BIND_PASSWORD: "secret-bind-user-password"
LDAP_USER_BASE: "ou=users,ou=archivebox,ou=services,dc=daniel-authenticator"
LDAP_USER_FILTER: "(objectClass=user)"

LDAP_USERNAME_ATTR: "uid"
LDAP_FIRSTNAME_ATTR: "givenName"
LDAP_LASTNAME_ATTR: "sn"
LDAP_EMAIL_ATTR: "mail"
```
This configuration works against [the LDAP server I use](https://github.com/DanielBatteryStapler/daniel-authenticator). There should be sufficient configuration options to work with any LDAP server, if properly configured.

The "LDAP" option defaults to False, in which case there are no changes to how Django handles authentication.

There is a bit of an annoyance with how this currently works, which is that when a user logins in for the first time, they are denied because they are not "staff". Once they get denied login for the first time, an existing superuser can mark them as "superuser" and then they will be able to login properly. I tried to figure out a way to set all LDAP users to superuser by default as a workaround, but I couldn't figure it out so that is not included in this PR. For now though, this behavior is okay for me personally because user authentication is still handled centrally, even if I have to specifically set all the users I want to have superuser permissions to actually be able to login. If in the future ArchiveBox does not require any special user permissions to login, then first-time logins will succeed.

# Related issues

#554 
Also I want all user authentication to go through LDAP on my personal server, so I went ahead and add that functionality.

# Changes these areas

- [ ] Bugfixes
- [X] Feature behavior
- [ ] Command line interface
- [X] Configuration options
- [ ] Internal architecture

Let me know if there is anything I can do to help get this merged.


pirate:
thanks for this work!

pirate:
The configuration is now documented here: https://github.com/ArchiveBox/ArchiveBox/wiki/Configuration#ldap

Issue #971, opened 2022-05-01
Bug: Parsing Wallabag RSS feed fails


peterrus:
#### Describe the bug
I have a setup where (through a cronjob) Archivebox fetches a RSS feed of my archived (aka read) articles from [Wallabag.it](https://wallabag.it) and imports them. This way I have a redundant archive of everything I read in Wallabag. Overkill? maybe.

Somewhere around 2022-03-31 the parsing of this RSS feed started to fail with the following error:

```
[+] [2022-05-01 10:03:42] Adding 5023 links to index (crawl depth=0)...
    > Saved verbatim input to sources/1651399422-import.txt
[X] Error while loading link! [1642338726.0] <link rel="via" "The Tragic Optimism of Software Projects"
Traceback (most recent call last):
[ ... redacted ...]
  File "/app/archivebox/index/schema.py", line 165, in typecheck
    assert isinstance(self.url, str) and '://' in self.url
AssertionError
```

Not long before 2022-03-31 Wallabag has released a new version: https://github.com/wallabag/wallabag/releases/tag/2.4.3 which includes a PR that modifies the formatting of the RSS feed it provides: https://github.com/wallabag/wallabag/pull/5347. I suspect this to be the culprit.

I am not exactly sure where the responsibility of fixing this lies but I want to at least document that I ran into this in case someone else experiences a similar issue.

#### Steps to reproduce

I have created a test account with one archived article on Wallabag.it. This account will expire in 14 days, but you can easily create a new one for testing purposes.

1. `curl https://app.wallabag.it/feed/dokafad/TDzxV9ejsZiWMq/archive | archivebox add --parser=wallabag_atom`

#### Screenshots or log output

Full log
```
[+] [2022-05-01 10:03:42] Adding 5023 links to index (crawl depth=0)...
    > Saved verbatim input to sources/1651399422-import.txt
[X] Error while loading link! [1642338726.0] <link rel="via" "The Tragic Optimism of Software Projects"
Traceback (most recent call last):
  File "/usr/local/bin/archivebox", line 33, in <module>
    sys.exit(load_entry_point('archivebox', 'console_scripts', 'archivebox')())
  File "/app/archivebox/cli/__init__.py", line 140, in main
    run_subcommand(
  File "/app/archivebox/cli/__init__.py", line 80, in run_subcommand
    module.main(args=subcommand_args, stdin=stdin, pwd=pwd)    # type: ignore
[...]

Issue #1000, opened 2022-07-18
Bug: Parsing Wallabag Atom feed tries to open nonexisting files


peterrus:
#### Describe the bug
After commit https://github.com/ArchiveBox/ArchiveBox/commit/a6767671fb68e25f67edcf16afafe5234d2826dd it seems that when `add()` (https://github.com/ArchiveBox/ArchiveBox/blob/a6767671fb68e25f67edcf16afafe5234d2826dd/archivebox/main.py#L573=) sometimes gets called with a `url` parameter containing the entire atom feed instead of a list of actual urls. I am not sure if this happens only with the Wallabag parser but this input is not expected at https://github.com/ArchiveBox/ArchiveBox/blob/a6767671fb68e25f67edcf16afafe5234d2826dd/archivebox/parsers/__init__.py#L158= and a `No such file or directory` error is raised.

#### Steps to reproduce
1. Run archivebox from any commit on the `dev` branch after https://github.com/ArchiveBox/ArchiveBox/commit/a6767671fb68e25f67edcf16afafe5234d2826dd
2. Import a wallabag feed, for example this test feed I created: `curl https://app.wallabag.it/feed/dokafad/TDzxV9ejsZiWMq/archive | archivebox add --parser=wallabag_atom`
3. The feeds seems to get imported, but a lot of errors are raised and a bunch of random paths get accessed (read) by the process.

#### Screenshots or log output

See: https://github.com/ArchiveBox/ArchiveBox/issues/971#issuecomment-1122499507=

#### ArchiveBox version

```logs
0.6.3
ArchiveBox v0.6.3 03eb7e5 Cpython Linux Linux-5.15.0-40-generic-x86_64-with-glibc2.31 x86_64
DEBUG=False IN_DOCKER=True IS_TTY=True TZ=UTC FS_ATOMIC=True FS_REMOTE=True FS_PERMS=644 999:999 SEARCH_BACKEND=ripgrep

[i] Dependency versions:
 √  PYTHON_BINARY         v3.10.5         valid     /usr/local/bin/python3.10                                                   
 √  SQLITE_BINARY         v2.6.0          valid     /usr/local/lib/python3.10/sqlite3/dbapi2.py                                 
 √  DJANGO_BINARY         v3.1.14         valid     /usr/local/lib/python3.10/site-packages/django/__init__.py                  
 √  ARCHIVEBOX_BINARY     v0.6.3          valid     /usr/local/bin/archivebox                                                   

[...]

PR #1215, opened 2023-08-20
Fix Wallabag Parser file access bug

peterrus:
# Summary

In this PR I attempt to debug https://github.com/ArchiveBox/ArchiveBox/issues/1000.

# Related issues

- Wallabag importing was fixed: https://github.com/ArchiveBox/ArchiveBox/issues/971

# Changes these areas

- [x] Temporary Workaround: Don't attempt to access filesystem in https://github.com/peterrus/ArchiveBox/blob/5b7efe256658dc30892a1d7f5c6d86a5b5ef0bc5/archivebox/parsers/__init__.py#L149

### DIFF
--- a/archivebox/parsers/__init__.py
+++ b/archivebox/parsers/__init__.py
@@ -149,17 +149,7 @@ def run_parser_functions(to_parse: IO[str], timer, root_url: Optional[str]=None,
 def save_text_as_source(raw_text: str, filename: str='{ts}-stdin.txt', out_dir: Path=OUTPUT_DIR) -> str:
     ts = str(datetime.now(timezone.utc).timestamp()).split('.', 1)[0]
     source_path = str(out_dir / SOURCES_DIR_NAME / filename.format(ts=ts))
-
-    referenced_texts = ''
-
-    for entry in raw_text.split():
-        try:
-            if Path(entry).exists():
-                referenced_texts += Path(entry).read_text()
-        except Exception as err:
-            print(err)
-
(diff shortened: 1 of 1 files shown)


peterrus:
@pirate I am not entirely sure what the purpose of the commit I reverted is, but it does seem to cause issues with Wallabag imports. `html`, `url_list` and `txt` parsers don't have any issues. So it must have something to do the massive amount of XML that comes out of the Wallabag feed. But even with e.g. the `html` parser `if Path(entry).exists():` tries to open (as in syscall) every line in the input, and I am not sure why you would want that because that would try to open things like `<html>` and `<a href="bla">link</a>` (which raises an error when lines are too long or contain invalid characters).

Is this just something we should remove?

pirate:
The removed code was checking for local file paths, e.g .if a user wants to import ~/Downloads/chrome_Bookmarks_export.txt. It's necessary for scheduled imports / other situations where you cant pipe a file in via stdin.

pirate:
Going to close this for now because that code is there intentionally and provides a valuable feature (loading URL lists from local paths). If you want to add additional error handling logic around it for a Wallabag issue please open a separate PR :)

peterrus:
Hi @pirate should we mark the Wallabag import as broken in the docs then? I have tried debugging the issue myself but I was unable to work something out as I seem to be missing a lot of context around Archivebox' architecture.

pirate:
Lets just keep #1000 open to track the Wallabag issue.

Issue #1181, opened 2023-07-14
Bug: Screenshots and DOM always fail after a while (Docker)


melyux:
#### Describe the bug

After running for a while and/or snapshotting a certain amount of URLs, the Chromium call that does screenshots and DOM starts failing. When I exec into the container and run the `chromium` command again manually, it always works. But when doing "Pull" from the web UI, it always fails. It only starts working again if I stop the container, `docker rm` it, and then start it again.

It's impossible from the log to see the exact Chromium error when this happens, because every single Chromium call is always prefixed by these error lines (and the ArchiveBox log only shows the first 5 lines):

```
find: ‘/root/.config/chromium/Crash Reports/pending/’: No such file or directory
[574:599:0714/201021.258704:ERROR:bus.cc(399)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[574:603:0714/201021.766940:ERROR:bus.cc(399)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[574:603:0714/201021.767034:ERROR:bus.cc(399)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[574:599:0714/201021.770660:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[574:599:0714/201021.771017:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[574:599:0714/201021.771115:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[574:599:0714/201021.771276:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[574:574:0714/201021.783527:ERROR:chrome_browser_cloud_management_controller.cc(162)] Cloud management controller initialization aborted as CBCM is not enabled.
[574:599:0714/201021.800785:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[...]

PR #1219, opened 2023-08-28
After a timeout, chrome will leave behind a SingletonLock, which prev…

spresse1:
# Summary

After a timeout in the pdf, screenshot, or dom extractor inside a docker container, chrome will leave behind a file at ~/.config/chromium/SingletonLock. This stops all three extractors from functioning until the docker container is completely torn down and regenerated.

This code adds removal of this file after a timeout _only when running in a docker container_. There is no behavior change outside a docker container. This was a deliberate choice on my part, as I don't want to interfere with a user's running chrome sessions when not running in a docker container.

# Related issues

#1181 

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/extractors/dom.py
+++ b/archivebox/extractors/dom.py
@@ -9,6 +9,7 @@ from ..util import (
     enforce_types,
     is_static_file,
     chrome_args,
+    chrome_cleanup,
 )
 from ..config import (
     TIMEOUT,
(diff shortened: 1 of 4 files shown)


pirate:
Thanks! This seems like a reasonable workaround for now until I finish the Huey + playwright event driven refactor.

PR #1241, opened 2023-10-09
Add new SQLite FTS5 full-text search backend

overhacked:
- WIP: add sqlite search backend boilerplate
- Introduce SQLite FTS5-powered search backend


pirate:
Great idea! I was thinking about doing this myself too but you beat me to it.

Have you tested to make sure the monkey patching at the top doesn't affect other areas of the codebase that depend on the native database connection provided by django?

archivebox/search/backends/sqlite.py
```diff
@@ -0,0 +1,164 @@
...
+    database = sqlite3.connect("search.sqlite3")
+    # Make connection callable, because `django.db.connection.cursor()`
+    # has to be called to get a context manager, but sqlite3.Connection
+    # is a context manager without being called.
+    def connection():
+        return database
+    SQLITE_BIND = "?"
+else:
+    from django.db import connection as database
+    connection = database.cursor
```
overhacked, line 26: Do you mean this line, when you say monkey patching? `connection` is a module-local symbol that doesn't affect `django.db`. Is there somewhere else I've fiddled with outside modules, or does the comment above just make it sound that way?

pirate: that's the line I meant, thanks for clarifying

I think I misread it as doing `django.db.connection = database.cursor` instead of just modifying module level local scope `connection = ...`.

pirate:
Let me know when you're ready for final review and I'll test/merge it!

overhacked:
I tested this on my "prod" dataset, and the index it created was **huge** at 1GB, 30% of the size of the `data/archive` folder. This seemed odd, because it's using FTS5 "contentless" indexes. I'm fairly sure that what's happening is that SQLite is indexing *very* long "terms", in the form of base64 strings in `data:` URLs, when it is given the contents of `singlefile.html` as the indexable content.

I thought about and experimented with several approaches to this, and I think the best approach is to (very loosely) parse `singlefile.html` and pass only text and some attributes to the search backend, rather than the entire HTML contents. I've opened a pull request with this approach: https://github.com/ArchiveBox/ArchiveBox/pull/1244. This resulted in a 10x reduction in index size. I think that merging this PR without somehow addressing the ballooning index size would result in issues down the road.

I think that Sonic doesn't have this problem because it limits maximum term length, but FTS5 doesn't have a configuration option for maximum (or minimum) term length. Or, maybe it's just the overall [`MAX_SONIC_TEXT_TOTAL_LENGTH`](https://github.com/ArchiveBox/ArchiveBox/blob/d7b883b04963673474f1f4d51ea5516eb767f326/archivebox/search/backends/sonic.py#L8)? Even if that's the case, the approach of parsing and including only meaningful text content, not JavaScript code, data URLs, and markup, would improve the signal-to-noise ratio of indexed content, so searching for `"html"` wouldn't hit on every document that had its `singlefile.html` indexed.

overhacked:
> Let me know when you're ready for final review and I'll test/merge it!

Alright, @pirate, I think it's ready for a hard look. Thanks!

pirate (approved):

Issue #1222, opened 2023-08-29
Bug: If document's title tag is empty, title extractor sets the snapshot title to "</title"


rmohns:
#### Describe the bug

I saved a webpage which is terribly coded by hand, and has an empty `title` tag. (Literally: `<title></title>'.) The resulting snapshot is named "</title". Easy to change but odd.

TBH I'm not sure if you should care, since we may not care if horribly invalid documents create errors. But on the off chance that it's easy to check for and change this in the code, am filing bug report. (Perhaps such snapshots could be named "No document title found".)

#### Steps to reproduce

1. Saved this page to ArchiveBox: http://wildwestcycle.com/f_oiltempdegradation.html
2. Snapshot title is `</title`

#### Screenshots or log output

![Screenshot 2023-08-29 at 5 05 12 PM](https://github.com/ArchiveBox/ArchiveBox/assets/2131133/ac6d6364-3177-467d-a27e-24e57589d806)

#### ArchiveBox version

```logs
ArchiveBox v0.6.2
Cpython Linux Linux-4.4.302+-x86_64-with-glibc2.28 x86_64
IN_DOCKER=True DEBUG=False IS_TTY=True TZ=UTC SEARCH_BACKEND_ENGINE=ripgrep

[i] Dependency versions:
 √  ARCHIVEBOX_BINARY     v0.6.2          valid     /usr/local/bin/archivebox                                                   
 √  PYTHON_BINARY         v3.9.5          valid     /usr/local/bin/python3.9                                                    
 √  DJANGO_BINARY         v3.1.10         valid     /usr/local/lib/python3.9/site-packages/django/bin/django-admin.py           
 √  CURL_BINARY           v7.64.0         valid     /usr/bin/curl                                                               
 √  WGET_BINARY           v1.20.1         valid     /usr/bin/wget                                                               
 √  NODE_BINARY           v15.14.0        valid     /usr/bin/node                                                               
 √  SINGLEFILE_BINARY     v0.3.16         valid     /node/node_modules/single-file/cli/single-file                              
 √  READABILITY_BINARY    v0.0.2          valid     /node/node_modules/readability-extractor/readability-extractor              
 √  MERCURY_BINARY        v1.0.0          valid     /node/node_modules/@postlight/mercury-parser/cli.js                         
 -  GIT_BINARY            -               disabled  /usr/bin/git                                                                
[...]

PR #1242, opened 2023-10-09
Fix HTML title parsing bugs.

benmuth:
# Summary

This slightly modifies the `HTML_TITLE_REGEX` and fixes two parsing errors. The first occurred when title tags were empty (e.g. `<title></title>`) which was parsed as `</title`. The second occurred when titles were a single character (e.g. `<title>A</title>`) which was not matched by the regex, and so the title would fall back to `link.base_url`.

With this change, now when tags are empty, the title falls back to `link.base_url`, and single character titles are parsed correctly.

I tested the new regex with the edge cases I could think of, and found some malformed HTML will still lead to undesired behavior. A more robust regex could probably be used in the future.

# Related issues

#1222 

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/extractors/title.py
+++ b/archivebox/extractors/title.py
@@ -26,7 +26,7 @@ from ..logging_util import TimedProgress
 
 HTML_TITLE_REGEX = re.compile(
     r'<title.*?>'                      # start matching text after <title> tag
-    r'(.[^<>]+)',                      # get everything up to these symbols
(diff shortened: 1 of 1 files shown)


pirate:
Thanks!

PR #1272, opened 2023-11-18
Add bookmarklet link back to add page

mikedamm:
It looks like bookmarklet link was accidently removed in 63c276a

# Summary

Adds missing bookmarklet link to add page template

# Related issues

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
I removed it because the [browser extension](https://github.com/tjhorner/archivebox-exporter) is a better way to link a browser, and also this page is only shown for a couple seconds now before an automatic redirect, so I don't think it's a good experience to ask the user to try and race to bookmark a link within 2 seconds! 😁

Issue #696, opened 2021-04-08
Show prompt to update in Web UI when ArchiveBox version is out of date


pirate:
Many people seem to be running older versions, which can lead to trouble when upgrading many versions at a time.

I'd like to nudge people towards upgrading more frequently by showing something in the UI.

Possibly using: https://github.com/releasepage/version.js

benmuth:
@pirate Couple of questions about the implementation for this:

- Should the majority of the logic be implemented in JS or Python? I've gotten something mostly working in JS, but I still have to implement the CLI hint which will require some Python modifications anyway.
- Should we cache the results of the GH releases API call for performance, or is that not important right now? I'm guessing the best way to do that would be to add or modify a Model?

pirate:
Ah just saw this, so unfortunately it has to be in Python not JS, because we need to show the warning in the CLI too, and I don't want every page request to phone home to GitHub so we'd want to do it once on startup of the server and just store it as a global config variable.

PR #1274, opened 2023-11-23
Show banner and hints to upgrade when ArchiveBox is out of date

benmuth:
# Summary
- adds a dismissable banner notifying admin users about newly available versions of ArchiveBox
- adds hints to upgrade to add and schedule jobs in the CLI

# Related issues
#696 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


archivebox/config.py
```diff
@@ -397,6 +398,55 @@ def get_commit_hash(config):
     except Exception:
         return None
 
+def get_version_releases(config):
+    """
+    returns a dictionary containing the GitHub release data for 
+    the recommended upgrade version and the currently installed version
+    """
+    github_releases_api = "https://api.github.com/repos/pirate/archivebox/releases"
```
pirate, line 406:
```suggestion
    github_releases_api = "https://api.github.com/repos/ArchiveBox/ArchiveBox/releases"
```

archivebox/config.py
```diff
@@ -397,6 +398,55 @@ def get_commit_hash(config):
...
+        stderr('Failed to get release data from GitHub', color='lightyellow', config=config)
+        return None
+
+    releases = response.json()
+    installed_version = config['VERSION']
+    installed_version_parts = parse_tag_name(installed_version)
+
+    # find current version or nearest older version (to link to)
+    current_version = None
+    for i, release in enumerate(releases):
```
pirate, line 418:
```suggestion
    for release in releases:
```

You can also use `... in releases.keys():` or `... in releases.values()` if it's a dictionary.

archivebox/config.py
```diff
@@ -696,6 +748,28 @@ def load_config(defaults: ConfigDefaultDict,
...
 #     with open(os.path.join(config['OUTPUT_DIR'], CONFIG_FILENAME), 'w+') as f:
 
+def parse_tag_name(v):
+    """parses a version tag string formatted like 'vx.x.x'"""
+    v = re.sub(r"\+.*$", "", v) # in case version string ends with '+editable'
+    parts = re.sub(r"^v", "", v).split(".")
+    return [int(p) for p in parts]
+
+
+def compare_versions(v1, v2):
```
pirate, line 758: We shouldn't need to do any manual version comparisons as the GitHub API should return the releases in order. Just use their position in the release list as the sort key. This will remain working if we move away from semver to date-based versions later.

archivebox/config.py
```diff
@@ -696,6 +748,28 @@ def load_config(defaults: ConfigDefaultDict,
 
 #     with open(os.path.join(config['OUTPUT_DIR'], CONFIG_FILENAME), 'w+') as f:
 
+def parse_tag_name(v):
+    """parses a version tag string formatted like 'vx.x.x'"""
+    v = re.sub(r"\+.*$", "", v) # in case version string ends with '+editable'
+    parts = re.sub(r"^v", "", v).split(".")
+    return [int(p) for p in parts]
```
pirate, line 755: Can be simplified to `(int(s) for s in v.split('+')[0].split('.'))`

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
         </div>
 
         <script>
+            if ("{{CAN_UPGRADE}}" === "True" && !localStorage.getItem("bannerDismissed")) {
+                let upgradeVersionTag = "{{VERSION_RELEASES.upgrade_version.tag_name}}"
```
pirate, line 146: These can all be `const`

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
...
+                banner.innerHTML = `
+                    <p>There's a new version of ArchiveBox available!</p>
+                    Your version: <a href=${currentVersionURL}>${currentVersionTag}</a> | New version: <a href=${upgradeVersionURL}>${upgradeVersionTag}</a>
+                    <p>
+                    <a href=https://github.com/ArchiveBox/ArchiveBox/wiki/Upgrading-or-Merging-Archives>Upgrading</a> | <a href=https://github.com/ArchiveBox/ArchiveBox/releases>Changelog</a> | <a href=https://github.com/ArchiveBox/ArchiveBox/wiki/Roadmap>Roadmap</a> 
+                    </p>
+                    <button id="dismiss-btn">Dismiss</button>
+                `
+                document.body.appendChild(banner);
+                let dismissButton = document.querySelector("#dismiss-btn")
```
pirate, line 166:
```suggestion
                const dismissButton = document.querySelector("#dismiss-btn")
```

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
...
+                let upgradeVersionTag = "{{VERSION_RELEASES.upgrade_version.tag_name}}"
+                let upgradeVersionURL= "{{VERSION_RELEASES.upgrade_version.html_url}}"
+                let currentVersionTag = "{{VERSION}}"
+                let currentVersionURL = "{{VERSION_RELEASES.current_version.html_url}}"
+
+                createBanner(currentVersionTag, currentVersionURL, upgradeVersionTag, upgradeVersionURL)
+            }
+
+            function createBanner(currentVersionTag, currentVersionURL, upgradeVersionTag, upgradeVersionURL){
+                const banner = document.createElement('div');
```
pirate, line 155:
```suggestion
                const banner = document.createElement('div')
```

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
         </div>
 
         <script>
+            if ("{{CAN_UPGRADE}}" === "True" && !localStorage.getItem("bannerDismissed")) {
+                let upgradeVersionTag = "{{VERSION_RELEASES.upgrade_version.tag_name}}"
+                let upgradeVersionURL= "{{VERSION_RELEASES.upgrade_version.html_url}}"
```
pirate, line 147:
```suggestion
                const upgradeVersionURL = "{{VERSION_RELEASES.upgrade_version.html_url}}"
```

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
...
+                let currentVersionTag = "{{VERSION}}"
+                let currentVersionURL = "{{VERSION_RELEASES.current_version.html_url}}"
+
+                createBanner(currentVersionTag, currentVersionURL, upgradeVersionTag, upgradeVersionURL)
+            }
+
+            function createBanner(currentVersionTag, currentVersionURL, upgradeVersionTag, upgradeVersionURL){
```
pirate, line 154:
```suggestion
            function createBanner(currentVersionTag, currentVersionURL, upgradeVersionTag, upgradeVersionURL) {
```

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
...
+                `
+                document.body.appendChild(banner);
+                let dismissButton = document.querySelector("#dismiss-btn")
+                if (dismissButton) {
+                    dismissButton.addEventListener("click", dismissBanner)
+                }
+            }
+
+            function dismissBanner() {
+                var banner = document.getElementById("upgrade-banner");
```
pirate, line 173:
```suggestion
                const banner = document.getElementById("upgrade-banner")
```

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
...
+                banner.style.display = "none";
```
pirate, line 174:
```suggestion
                banner.style.display = "none"
```

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
...
+                localStorage.setItem("bannerDismissed", "true");
```
pirate, line 175:
```suggestion
                localStorage.setItem("bannerDismissed", "true")
```

archivebox/templates/admin/base.html
```diff
@@ -123,6 +142,39 @@ <h1 id="site-name">
         </div>
 
         <script>
+            if ("{{CAN_UPGRADE}}" === "True" && !localStorage.getItem("bannerDismissed")) {
```
pirate, line 145: Can you also check to make sure user is logged in here

pirate (changes requested):
Looks good, just some small comments

archivebox/config.py
benmuth: The only thing I thought this is needed for is to link to the release page for the currently installed version. I thought a comparison is necessary in case the installed version name doesn't match any of the version names in the releases (like if the installed version is "0.7.0"). I decided to link to the first version in the list of releases that's older than the currently installed, but maybe we shouldn't link to anything in this case?

benmuth: I removed that function, now just making the parsing function return a tuple which can be compared directly (I think this is basically what you suggested in another comment). If we change to date-based versions I think the logic can stay the same.

pirate:
made some changes, take a look at my commits

Issue #1304, opened 2023-12-30
Bug: ArchiveBox doesn't work on NFS/SMB/FUSE drives that disallow root ownership or have `root_squash` set


gnattu:
#### Describe the bug

If the docker-entry point failed to chown the archive folder, it quits immediately, so that the container does not run at all, make it impossible to run archivebox with an nfs-mounted volume as the archive folder.

I'm currently using this with a modified entrypoint which removed the line to chmod everything under DATA dir.

#### Steps to reproduce

1. Create an NFS volume with docker: `docker volume create --driver local --opt type=nfs  --opt o=addr=[ip-address],rw  --opt device=:[path-to-directory] [volume-name]`

2. Use this as the archivefolder in docker-compose.yml: `- volume-name:/data/archive`

3. Start container. The conatiner quit right after it failed to chown the nfs mounted folder

#### Screenshots or log output

```
[+] Creating 1/0
 ✔ Container archivebox-sonic-1  Running                                                                     0.0s
chown: changing ownership of '/data/archive': Operation not permitted
```

#### ArchiveBox version

```logs
[+] Creating 1/0
 ✔ Container archivebox-sonic-1  Running                                                                     0.0s
0.7.1
ArchiveBox v0.7.1+editable BUILD_TIME=2023-12-18 06:57:08 1702882628
IN_DOCKER=True IN_QEMU=False ARCH=aarch64 OS=Linux PLATFORM=Linux-6.5.13-orbstack-00121-ge428743e4e98-aarch64-with-glibc2.36 PYTHON=Cpython
FS_ATOMIC=True FS_REMOTE=True FS_USER=1000:1000 FS_PERMS=644
DEBUG=False IS_TTY=True TZ=UTC SEARCH_BACKEND=sonic LDAP=False

[i] Dependency versions:
 √  PYTHON_BINARY         v3.11.7         valid     /usr/local/bin/python3.11                                     
 √  SQLITE_BINARY         v2.6.0          valid     /usr/local/lib/python3.11/sqlite3/dbapi2.py                   
 √  DJANGO_BINARY         v3.1.14         valid     /usr/local/lib/python3.11/site-packages/django/__init__.py    
 √  ARCHIVEBOX_BINARY     v0.7.1          valid     /usr/local/bin/archivebox                                     

 √  CURL_BINARY           v8.4.0          valid     /usr/bin/curl                                                 
[...]

PR #1312, opened 2024-01-05
fix: handle archive folder permission more graceful

gnattu:
# Summary

On mounted network shares, `chmod` might be hard or impossible to do without modify the server config. However, we can ignore the archive folder because we already tested if that folder is writeable and printed warnings to the user when we failed.

This PR ignores the archive folder and its contents when doing the chmod step, which is useful for mounted NFS/CIFS shares on archive folder specifically but the rest of the data folder remains a local directory. When modifying the config of the server is not an option, this can still enable users to use archivebox with network shares to store the most storage consuming part of the web archives. 

This PR also modifies the warning message to be more accurately reflect the permission error is happening on the archive folder specifically, not the whole data folder.

I also noticed that there are "create archive folder if not exist" comment, but the shell script actually create the logs folder. I don't know if it is the intended behavior so I leave it as-is.

# Related issues

This is a followup of #1304, providing an alternative method when modifying server config is not an option.

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/bin/docker_entrypoint.sh
+++ b/bin/docker_entrypoint.sh
@@ -35,7 +35,7 @@ export DEFAULT_PGID=911
 if [[ "$PUID" == "0" ]]; then
     echo -e "\n[X] Error: Got PUID=$PUID and PGID=$PGID but ArchiveBox is not allowed to be run as root, please change or unset PUID & PGID and try again." > /dev/stderr
     echo -e "    Hint: some NFS/SMB/FUSE/etc. filesystems force-remap/ignore all permissions," > /dev/stderr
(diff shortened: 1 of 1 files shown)


pirate:
I still want to try chowning the entire `./data` folder first, since the best-case/most common scenario is that everything inside is owned by PUID:PGID with the same permissions.

I pushed a commit to slightly modify your code to be a fallback if a full `chown` fails.

pirate (approved):

PR #1314, opened 2024-01-09
Show the upgrade notification only in specific views

benmuth:
# Summary

This PR adds back the upgrade notification banner that was originally added in #1274 and removed in 4af9bee, but only to specific views instead of the entire admin site, as discussed in #1306.

I wasn't sure for which views we should show the banner, so for now it's only showing on the snapshot view (the page at `/admin/core/snapshot/`.

The necessary context is provided by overriding the `changelist_view` method of the `SnapshotAdmin` model. I think this could work for all views except for the main admin page (`/admin/`). If we want to pass context to the view for that page, I think we can override the `each_context` method of `ArchiveBoxAdmin` instead.

# Related issues
#696 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
Looks good. Can you post a screenshot of what this looks like in action?

benmuth:
Here's what it looks like for me: image

pirate:
Awesome, thanks.

Issue #1316, opened 2024-01-11
Text Search and Filters don't work at the same time in the web UI


sbutcher:
#### Describe the bug

In the UI, providing a search term, then applying a filter (e.g. changed today, or a tag) does not make any difference to the results. I would expect to see the filter applied to the results on screen.

#### Steps to reproduce

Running archivebox dev version
Open the web UI and enter a search term that gives results e.g. football
Then click the "Today" option in the "By date added" in the filter column

#### Screenshots or log output

![image](https://github.com/ArchiveBox/ArchiveBox/assets/4619907/f9cecaea-203c-45f7-97e2-0b87fb1c4afb)

clicking the "Today" in the filter view makes no change to the results, where the expected result is to show a subset of the results, combining all results with teh word football that were added today. The same issue applies to any filters while there is a search term specified. If you clear the search term and show all results, then the filters work correctly

#### ArchiveBox version
```
0.7.2
ArchiveBox v0.7.2+editable COMMIT_HASH=e43babb BUILD_TIME=2024-01-06 01:22:06 1704504126
IN_DOCKER=True IN_QEMU=False ARCH=x86_64 OS=Linux PLATFORM=Linux-6.2.0-39-generic-x86_64-with-glibc2.36 PYTHON=Cpython
FS_ATOMIC=True FS_REMOTE=True FS_USER=911:911 FS_PERMS=644
DEBUG=False IS_TTY=True TZ=UTC SEARCH_BACKEND=ripgrep LDAP=False

[i] Dependency versions:
 √  PYTHON_BINARY         v3.11.7         valid     /usr/local/bin/python3.11                                                   
 √  SQLITE_BINARY         v2.6.0          valid     /usr/local/lib/python3.11/sqlite3/dbapi2.py                                 
 √  DJANGO_BINARY         v3.1.14         valid     /usr/local/lib/python3.11/site-packages/django/__init__.py                  
 √  ARCHIVEBOX_BINARY     v0.7.2          valid     /usr/local/bin/archivebox                                                   

 √  CURL_BINARY           v8.5.0          valid     /usr/bin/curl                                                               
 √  WGET_BINARY           v1.21.3         valid     /usr/bin/wget                                                               
 √  NODE_BINARY           v20.10.0        valid     /usr/bin/node                                                               
 √  SINGLEFILE_BINARY     v1.1.46         valid     /app/node_modules/single-file-cli/single-file                               
[...]

PR #1333, opened 2024-01-21
Text Search and Filters don't work at the same time in the web UI #1316

neel-suthar:
Making sure to return distinct results. Changing set operation to '&' to show the matching results from filters AND search term

Fixes: https://github.com/ArchiveBox/ArchiveBox/issues/1316

# Summary

The fix involves adjusting the combination of filter and search results to avoid duplicates. Instead of using the "or" (|) operation, consider using a method that ensures the snapshots appearing in the result set satisfy both the filter conditions and the search term. This adjustment aims to eliminate duplicate entries in the combined result.

Before the fix:
image

After the fix:
image

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


neel-suthar:
@pirate Please review.

pirate:
I'm worried that while this enables filters working, it will hide some results that only match via the search index but dont match via db field search.

e.g. if user searches for some string that's only present in a tag, but not in the snapshot content

Maybe better to extract the filter parameters passed from the UI and apply them to the queryset manually?

neel-suthar:
Interesting. I thought the search results would contain that information. So when we do perform & operation the final set would have the results matching both. Let me go through the code and figure it out. Thanks.

archivebox/core/mixins.py
```diff
@@ -10,10 +10,10 @@ def get_search_results(self, request, queryset, search_term: str):
 
         search_term = search_term.strip()
         if not search_term:
-            return qs, use_distinct
+            return qs.distinct(), use_distinct
         try:
             qsearch = query_search_index(search_term)
-            qs = qs | qsearch
+            qs = qs & qsearch
```
pirate, line 16:
```suggestion
            qs = qs | qsearch
```

reverting just this change for now so I can merge the `.distinct()` fix

PR #1335, opened 2024-01-24
ability to populate is_staff and is_superuser flags at LDAP authentication

vladimirdulov:
On LDAP authentication, it sets `is_staff` flag to `True` for all LDAP users and sets` is_superuser` to `True` if `LDAP_CREATE_SUPERUSER` config setting is set to True accordingly.

It's needed to make LDAP auth properly working.


archivebox/core/auth_ldap.py
```diff
@@ -0,0 +1,11 @@
...
+from ..config import (
+    LDAP_CREATE_SUPERUSER
+)
+
+def create_user(sender, user=None, ldap_user=None, **kwargs):
+
+    if not user.id and LDAP_CREATE_SUPERUSER:
+        user.is_superuser = True
+
+    user.is_staff = True
```
pirate, line 11:
```suggestion
    user.is_staff = True
    print(f'[!] WARNING: Creating new user {user} based on LDAP user {ldap_user} (is_staff={user.is_staff}, is_superuser={user.is_superuser})')
```

pirate:
thanks!

Issue #1025, opened 2022-09-12
Footguns: `WGET_ARGS`, `CURL_ARGS`, SINGLEFILE_ARGS


notevenaperson:
These options let the user shoot themselves in the foot, silently. I think the documentation or the variables themselves should be changed to be more ergonomic.

Did you need to add a header to wget? 
```py
WGET_ARGS=['--header=Accept-Language: en-US,en']
```
Look at this configuration, it looks inoffensive. Checking [the documention](https://github.com/ArchiveBox/ArchiveBox/wiki/Configuration#wget_args), nothing ticks you off you're using it wrong. **But no.**
After this, your 🆆 button won't take you to the main html you archived, and wget archives things slighlty differently, silently. 

Why? By setting WGET_ARGS you overwrote [vital settings](https://github.com/ArchiveBox/ArchiveBox/blob/dev/archivebox/config.py#L159), as it turns out they're also stored in WGET_ARGS. The documentation doesn't tell you about this. This happened to me, I just happily overwrote the variable.

Proposal:
- [ ] The documentation for `*_ARGS` should have a good warning or display the default value, so that users can suspect they're overwriting something.
- [ ] Or, create `EXTRA_WGET_ARGS`, `EXTRA_CURL_ARGS`, and so on. These won't overwrite the now-considered low-level `*_ARGS`. `EXTRA_*_ARGS` shall be the more user-facing option and more promoted in the documentation.

pirate:
I'm ok with the EXTRA_ options as I want users to be able to override the defaults.

Issue #990, opened 2022-06-16
Feature Request: .md extractor


cmuc24:
## Feature Request, a short description related to extractors, result items
As a admin, i need a markdown format, similar to the existing readability format, 
to catch .md  items for further, external downstream processes 

## the ideal specific solution
As a user, i can define MD as (additional) format for my (to be archived) items, 
to get these items in our publishing worklow and as a result, 
back to our commonly used MD based knowledge base.

Good for Applications like MkDocs, Trilium or similar other MD based Applications.

## actual solution
At the moment a MD Webclipper as browser extension do the job, with some additional options also.
https://github.com/deathau/markdownload  (MIT and Apache2 licenses only)

## how important 
as i can (not clear) see, a twin from the readability extractor, paired with great stuff, e.g. from mentioned .md Webclipper,
could have potential for a low hanging fruit!? :-)

**personally** i'll say its an improvement that can increase (not only) effiecenty in downstream processes 

 - [x] I'm willing to contribute [dev time](https://github.com/ArchiveBox/ArchiveBox#archivebox-development) / [money](https://github.com/sponsors/pirate) to fix this issue
 - [x] I like ArchiveBox so far / would recommend it to a friend
 - [x] I've had a lot of difficulty getting ArchiveBox set up (i´ll describe it outside this feature request)

notevenaperson:
The mercury extractor has [--format=markdown](https://github.com/postlight/parser/issues/316). In archivebox the option is set to [--format=text](https://github.com/ArchiveBox/ArchiveBox/blob/dev/archivebox/extractors/mercury.py#L68) You just need a way to change the argument. 

(`MERCURY_ARGS` config option when? 😄 (I could work on that actually)

PR #1360, opened 2024-02-23
Add `_EXTRA_ARGS` for various extractors

benmuth:
# Summary
This PR adds a way to configure `wget`, `curl`, `singlefile`, `youtube-dl`, and `chrome` without overriding the default options.

The main default options, extra options, and more specific options (like `WGET_USER_AGENT`)  are all deduplicated. It's assumed that options set with more specificity should take precedence, so something like the `--user-agent` argument for `wget` will come from `WGET_USER_AGENT` instead of `_ARGS` or `_EXTRA_ARGS`, and options set in `_EXTRA_ARGS` take precedence over `_ARGS`.

This PR might need some more testing with more complex configurations. Hopefully it's simple enough that won't break anything while still being useful, but I'm not a wizard with `curl` or `wget` so there might be some possibilities I don't know about.

# Related issues
#1025 

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/config.py
+++ b/archivebox/config.py
@@ -176,6 +176,7 @@ CONFIG_SCHEMA: Dict[str, ConfigDefaultDict] = {
                                                                 '--add-metadata',
                                                                 '--format=(bv*+ba/b)[filesize<={}][filesize_approx<=?{}]/(bv*+ba/b)'.format(c['MEDIA_MAX_SIZE'], c['MEDIA_MAX_SIZE']),
                                                                 ]},
+        'YOUTUBEDL_EXTRA_ARGS':     {'type': list,  'default': None},
 
 
         'WGET_ARGS':                {'type': list,  'default': ['--no-verbose',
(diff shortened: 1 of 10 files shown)


pirate:
Awesome! thanks for doing this, it's a great addition

should we add `CHROME_EXTRA_ARGS` and `YOUTUBEDL_EXTRA_ARGS` too?

pirate:
Looks great!

One minor change then I'm ready to merge:

Can you flip the `dedupe()` precedence order to last arg wins?

I most CLI tools that accept duplicate args take the *last* arg as the winning one if they see the same arg twice, so I want to follow that convention.

pirate:
Related: #990

benmuth:
> Related:
> 
> * [Feature Request: Add `MERCURY_ARGS` extractor option to enable saving article text as `.md` markdown #990](https://github.com/ArchiveBox/ArchiveBox/issues/990)

Oh cool, I'll take a look at that next.

benmuth:
@pirate I went ahead and added both the `ARGS` and `EXTRA_ARGS` options for the Mercury extractor to be more consistent with the rest of the changes in this PR.

Issue #761, opened 2021-06-05
Bug: Title becomes "Before you continue to YouTube" when archiving youtube videos


dansbandit:
#### Describe the bug

Title becomes 'Before you continue to YouTube' instead of video title due to youtube redirects to a cookie consent form. This could be solved if you could add a cookie file to the curl command that is run.
```
["curl", "--silent", "--location", "--compressed", "--max-time", "60", "--user-agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.61 Safari/537.36 ArchiveBox/0.6.2 (+https://github.com/ArchiveBox/ArchiveBox/) curl/curl 7.76.0 (amd64-portbld-freebsd12.2)", "https://www.youtube.com/watch?v=aP8sRCun63M"]```

#### Steps to reproduce

1. `archivebox add https://www.youtube.com/watch?v=aP8sRCun63M`
2. Title becomes 'Before you continue to YouTube' when it should be 'ArchiveBox'

#### Screenshots or log output

N/A

#### ArchiveBox version
```
ArchiveBox v0.6.2
Cpython FreeBSD FreeBSD-12.2-RELEASE-p6-amd64-64bit-ELF amd64
IN_DOCKER=False DEBUG=False IS_TTY=True TZ=UTC SEARCH_BACKEND_ENGINE=ripgrep

[i] Dependency versions:
 √  ARCHIVEBOX_BINARY     v0.6.2          valid     ./.local/bin/archivebox                                                     
 √  PYTHON_BINARY         v3.7.10         valid     /usr/local/bin/python3.7                                                    
 √  DJANGO_BINARY         v3.1.12         valid     ./.local/lib/python3.7/site-packages/django/bin/django-admin.py             
 √  CURL_BINARY           v7.76.0         valid     /usr/local/bin/curl                                                         
 √  WGET_BINARY           v1.21           valid     /usr/local/bin/wget                                                         
 √  NODE_BINARY           v14.16.1        valid     /usr/local/bin/node                                                         
 √  SINGLEFILE_BINARY     v0.3.13         valid     ./node_modules/single-file/cli/single-file                                  
 √  READABILITY_BINARY    v0.1.0          valid     ./node_modules/readability-extractor/readability-extractor                  
[...]

PR #1364, opened 2024-02-27
Use `COOKIES_FILE` to fetch page titles

benmuth:
# Summary
This PR lets the title extractor use the `COOKIES_FILE`, if available. This helps avoid extracting the titles of captcha or login pages.

# Related issues
#761 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [x] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/util.py
+++ b/archivebox/util.py
@@ -3,6 +3,7 @@ __package__ = 'archivebox'
 import re
 import requests
 import json as pyjson
+import http.cookiejar
 
 from typing import List, Optional, Any
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


archivebox/util.py
```diff
@@ -164,13 +165,26 @@ def parse_date(date: Any) -> Optional[datetime]:
...
+    from .config import (
+         TIMEOUT,
+         CHECK_SSL_VALIDITY,
+         WGET_USER_AGENT,
+         COOKIES_FILE,
+    )
     timeout = timeout or TIMEOUT
+
+    cookie_jar = http.cookiejar.MozillaCookieJar()
+    if COOKIES_FILE is not None:
```
pirate, line 177:
```suggestion
    if COOKIES_FILE and Path(COOKIES_FILE).is_file():
```

pirate:
Down to merge this if you've tested it! Just change it to `Ready for Review` when you're ready :)

benmuth:
I've been testing it and I'm actually having trouble verifying that this works. I thought it was working when I opened this PR, but it seems I was wrong. For my test case, the title isn't correct if the title extractor is the only enabled extractor. It is correct when `wget` is enabled, so the cookie file is being read correctly by ArchiveBox, but my fix here doesn't seem to be working. Not sure where it's failing yet.

benmuth:
After throwing some stuff at the wall, I got to a solution that works for me. For some reason, the cookie file only seems to work with `requests` when I set each cookie one by one, like 

```
        for cookie in cookie_jar:
            session.cookies.set(cookie.name, cookie.value, domain=cookie.domain, path=cookie.path)
```
It didn't work for me when I tried `session.cookies = cookie_jar` or `session.cookies.update(cookie_jar)`, and from the docs I can't tell why. If anyone could tell me what I'm missing, I'd appreciate it.

PR #1365, opened 2024-02-27
Fix generic_json parser

### LINKED ISSUE #1347: Bug: `JSONDecodeError` when trying to load JSON with an array at the top-level
With ArchiveBox version 0.6.2, I used to import URLs stored in JSON files with content looking like the following:
```json
[
    {"url": "https://archivebox.io/", "tags": "test,archivebox"},
    {"url": "https://en.wikipedia.org/wiki/Linux", "tags": "test,wikipedia"}
]
```
```bash
archivebox add --parser json < ./links.json
```
Everything worked well.

With version 0.7.2, however, I have a `JSONDecodeError` exception during the import:
```
  File "/tmp/archivebox/venv/lib/python3.11/site-packages/archivebox/main.py", line 631, in add
    new_links += parse_links_from_source(write_ahead_log, root_url=None, parser=parser)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/archivebox/venv/lib/python3.11/site-packages/archivebox/util.py", line 116, in typechecked_function
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/archivebox/venv/lib/python3.11/site-packages/archivebox/index/__init__.py", line 278, in parse_links_from_source
    raw_links, parser_name = parse_links(source_path, root_url=root_url, parser=parser)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/archivebox/venv/lib/python3.11/site-packages/archivebox/util.py", line 116, in typechecked_function
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/archivebox/venv/lib/python3.11/site-packages/archivebox/parsers/__init__.py", line 103, in parse_links
    links, parser = run_parser_functions(file, timer, root_url=root_url, parser=parser)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/archivebox/venv/lib/python3.11/site-packages/archivebox/parsers/__init__.py", line 117, in run_parser_functions
    parsed_links = list(parser_func(to_parse, root_url=root_url))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/archivebox/venv/lib/python3.11/site-packages/archivebox/parsers/generic_json.py", line 23, in parse_generic_json_export
    links = json.loads(json_file_json_str)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/decoder.py", line 340, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 1 column 61 (char 60)
```
The error is caused by the following line, introduced by commit aaca74f: https://github.com/ArchiveBox/ArchiveBox/blob/3ad32509e985236f82f3558f31b856623b1eb261/archivebox/parsers/generic_json.py#L22

jimwins:
# Summary

This fixes the `generic_json` parser by not always assuming the JSON needs special handling and doing a more straightforward workaround when it might.

Also adds support for a tags field.

# Related issues

Fixes #1347.

# Changes these areas

- [X] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/parsers/generic_json.py
+++ b/archivebox/parsers/generic_json.py
@@ -66,9 +66,14 @@ def parse_generic_json_export(json_file: IO[str], **_kwargs) -> Iterable[Link]:
             elif link.get('name'):
                 title = link['name'].strip()
 
-            tags = ''
-            if link.get('tags'):
-                tags = link.get('tags').replace(' ',',')
+            # if we have a list, join it with commas
+            tags = link.get('tags')
+            if type(tags) == list:
+                tags = ','.join(tags)
+            elif type(tags) == str:
+                # if there's no comma, assume it was space-separated
+                if ',' not in tags:
+                    tags = tags.replace(' ', ',')
 
             yield Link(
                 url=htmldecode(url),
--- a/archivebox/parsers/generic_json.py
+++ b/archivebox/parsers/generic_json.py
@@ -18,9 +18,16 @@ def parse_generic_json_export(json_file: IO[str], **_kwargs) -> Iterable[Link]:
 
     json_file.seek(0)
 
-    # sometimes the first line is a comment or filepath, so we get everything after the first {
-    json_file_json_str = '{' + json_file.read().split('{', 1)[-1]
-    links = json.loads(json_file_json_str)
+    try:
+        links = json.load(json_file)
+    except json.decoder.JSONDecodeError:
+        # sometimes the first line is a comment or other junk, so try without
+        json_file.seek(0)
+        first_line = json_file.readline()
+        #print('      > Trying JSON parser without first line: "', first_line.strip(), '"', sep= '')
+        links = json.load(json_file)
+        # we may fail again, which means we really don't know what to do
+
     json_date = lambda s: datetime.strptime(s, '%Y-%m-%dT%H:%M:%S%z')
 
     for link in links:
@@ -59,11 +66,15 @@ def parse_generic_json_export(json_file: IO[str], **_kwargs) -> Iterable[Link]:
             elif link.get('name'):
                 title = link['name'].strip()
 
+            tags = ''
+            if link.get('tags'):
+                tags = link.get('tags').replace(' ',',')
+
             yield Link(
                 url=htmldecode(url),
                 timestamp=ts_str,
                 title=htmldecode(title) or None,
-                tags=htmldecode(link.get('tags')) or '',
+                tags=htmldecode(tags),
                 sources=[json_file.name],
             )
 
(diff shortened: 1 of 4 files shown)


archivebox/parsers/generic_json.py
```diff
@@ -59,11 +66,15 @@ def parse_generic_json_export(json_file: IO[str], **_kwargs) -> Iterable[Link]:
             elif link.get('name'):
                 title = link['name'].strip()
 
+            tags = ''
+            if link.get('tags'):
+                tags = link.get('tags').replace(' ',',')
```
pirate, line 71: we do support tags containing spaces (but not newlines), so it might be better to replace '\n' with ','

jimwins: I was working off the example line and what Pinboard's JSON export looks like, which has space-separated tags. A more ideal JSON file would probably have `"tags": [ "tag1", "tag2", "tag with space" ]`, so maybe this should check to see `link.get('tags')` gives us a list, but then we have to put it together with commas because `Link()` can't handle `tags` as a list yet. I figure this could be revisited when that's tackled from the Roadmap.

Cleaning the `\n` out of tags should probably happen at a lower level, otherwise it will just be duplicated in every parser or other place tags enter the system.

More test data based on real-world examples would probably be good.

pirate: Yeah that sounds good, lets try to check if it's a comma-separated string, space-separated string, or array. I have seen all 3 in the wild.

And I agree we should definitely collect more parser test inputs and add them to `tests/`.

jimwins: Okay, new commit coming. Handles lists, assumes a single string with any commas is already comma-separated, and anything else is space-separated so we turn those into commas. The edge case we'll mishandle is a JSON file that assumes the single-string version of `tag` is comma-separated and an entry only has one tag with a space in it. It probably makes the most sense to just get that wrong until someone provides evidence a tool that actually produces JSON files like that.

Maybe we could even be more clever by using `TAG_SEPARATOR_PATTERN` but that starts to feel like going overboard and I have a feeling that is a whole can of worms that nobody should open.

pirate: Yup! My thoughts exactly, that edge case is probably rare.

Thanks for doing all this btw! This is a great fix / improvement to a fairly stale and brittle area of the codebase.

pirate (approved):

pirate:
Looks good, thanks! Ready to merge @jimwins?

jimwins:
Yeah, I think this is good to go. I'll open a new issue to track adding JSONL handling.

Issue #1126, opened 2023-03-22
Discussion: Sponsoring the REST API development + applying for grants to fund ArchiveBox development


djkemmet:
I would like to sponsor (pay a developer cash monies in USD) for the development of the API for ArchiveBox. I need this for my digital preservation and ML/NLP work and I suspect a lot of people will end up doing a lot of cool things with it. I would like to talk to @pirate  and get his blessing and solicit feedback from the community on how we could work together to accomplish this as well as what a reasonable rate and time line would be. @pirate  is this something you have the bandwidth and inclination to do? if not could you recommend anyone as tribute? :P 

Ideally I see the community coming to consensus on what an initial API would look like in the form of an Issue, that would serve as the scope of work for the commission, a $ amount would be set against that issue and once complete and the PR is accepted the money would be paid to the developer from money in escrow for this task. 

Thank you all for your time with this and I look forward to hearing from you.

pirate:
Hey @djkemmet, this is definitely something I'm interested in and I'm excited that you're willing to contribute towards seeing it happen.

I'm getting ready to write some grant applications for ArchiveBox and I'm hoping to fund part-time development for the next few years. The API is definitely one of the main things I want to build in the next year, and I want to work on ArchiveBox as my main job if I can pull off finding a big sponsor.

The grants I'm aiming for are $150k+ so if your budget for this task is <$2k I wouldn't bother sponsoring this if you cant wait a few months, as I'll likely get around to it anyway. If your needs are more urgent than that or your budget is higher then lets discuss further via chat/call: https://calendly.com/nicksweeting/choose-a-time/

I agree either way we should finalize an API spec and framework choice, the work so far toward that is here: https://github.com/ArchiveBox/ArchiveBox/issues/496#issuecomment-1097310582

Either way you can comment your thoughts on that spec in that issue, and we can continue the discussion about timeline/sponsorship/and your specific needs here.

aaronmayer108:
WOO! So excited to see that grant money might be a possibility! Btw, grant writing is becoming trivially easy with GPT.

djkemmet:
heyy @pirate , 

yeah my budget for this was $1000 USD If there's anyone in your network who might be suited for this work and assuming 1000$ is a resonable amount of money to get the API off the ground I would definitely be happy to sponsor someone based on your recommendation. that said, as far as grants go, have you looked at futo? https://futo.org/fellows/

pe1uca:
Hello guys!  
[...]

PR #1327, opened 2024-01-17
OLD IDEA / no longer active: New modular plugin design + 3 new plugins: browsertrix-crawler, ReplayWeb.page, gallery-dl

pirate:
Coming soon:
- Ability to enable/disable/configure ArchiveBox submodules via the admin UI

image
image
image

PR #1397, opened 2024-04-09
REST API v1 using `django-ninja`

Brandl:
# Summary

I recently integrated Archivebox with a small python script. The current approach of setting up the full django environment is somewhat cumbersome and limits usecases where the interaction is triggered by remote systems/containers.

So here goes my initial draft of a REST API based on django ninja.

For Auth I implemented Token/Bearer Auth. Token can be obtained and validated via API Endpoints.
Ideally there should be a UI (Django Admin,...) to make obtaining them more friendly to users.

The current API resembles the CLI for compatibility and simplicity. 
Adding CRUD Endpoints should be rather straight forward.

Ideally Endpoints should return a TASK-ID for long running tasks for long polling or allow for SSE subscription.
This would require to update to Django 4.2 for async support.

# Related issues

Touches on all the API issues:
#496 
#1126 

# Changes these areas

- [X] Token model
- [x] Token based authentication
- [X] ADD/UPDATE/REMOVE/LIST Endpoints
- [ ] CRUD Endpoints for Snapshot/Link,...
- [ ] UI for Token 
- [ ] Extensive Testing suite
- [ ] (future) streaming/fully async endpoints 

Screenshot from 2024-04-10 01-45-11


pirate:
Awesome work, thanks for doing this!

I'll dig into it next week, I think it should be fairly straightforward to merge.


> I think allowing the user to create named API tokens (that inherit the same permissions as that user) is enough for now, no need to do OAuth2 or JWT

100% agree, sounds good.

> Ideally there should be a UI (Django Admin,...) to make obtaining them more friendly to users.

I've been using `django-solo` to quickly expose config options via the Django Admin in my other branch: https://github.com/ArchiveBox/ArchiveBox/pull/1327.  I think we can add a pane to manage API tokens without too much trouble.

> Ideally Endpoints should return a TASK-ID for long running tasks for long polling or allow for SSE subscription.

I'm planning on integrating `django-huey-monitor` in the future, it would provide task IDs and manage TaskResult rows, I think for now we can just return the Snapshot primary key.

> This would require to update to Django 4.2 for async support.

We are actually on Django 4.2 already on `dev` as of recently (see [v0.8.0-rc](https://github.com/ArchiveBox/ArchiveBox/releases/tag/v0.8.0-rc) alpha release notes).

pirate:
I made a bunch of changes on `dev` FYI: 7515325

Screenshot 2024-04-25 at 3 56 38 AM

Issue #720, opened 2021-04-24
Feature Request: Extract DOI numbers and download academic paper PDFs automatically


pirate:
New extractor idea: SCIHUB

e.g. take this academic paper for example: https://www.cell.com/current-biology/fulltext/S0960-9822(19)31469-1

If a full paper PDF is available on scihub, e.g.: https://sci-hub.se/https://www.cell.com/current-biology/fulltext/S0960-9822(19)31469-1 it could be downloaded to a `./archive/<timestmap>/scihub/` output folder.

```bash
# try downloading via verbatim URL first
$ scihub.py -d https://www.cell.com/current-biology/fulltext/S0960-9822(19)31469-1'
DEBUG:Sci-Hub:Successfully downloaded file with identifier https://www.cell.com/current-biology/fulltext/S0960-9822(19)31469-1
```

We could also look for a DOI number in the page URL or page html contents e.g.: `10.1016/j.cub.2019.11.030` using a regex and try downloading that.
```bash
# otherwise try downloading via any regex-extracted bare DOI numbers on the page or in the URL
$ scihub.py -d '10.1016/j.cub.2019.11.030'
DEBUG:Sci-Hub:Successfully downloaded file with identifier 10.1016/j.cub.2019.11.030

$ ls
c28dc1242df6f931c29b9cd445a55597-.cub.2019.11.030.pdf
```

**New Dependencies:**
- https://github.com/zaytoun/scihub.py

**New Extractors:**
- `extractors/scihub.py`

**New Config Options:**
- `SAVE_SCIHUB=True`

lbcnz:
Captchas would be a problem.

pirate:
In my testing so far it's not a problem as long as you're not doing dozens of papers at a time.

jezcope:
Two more avenues to explore:

1. Zotero has a big collection of translators to get full text from many different publishers and repositories https://github.com/zotero/translators
2. Unpaywall has a huge database of open access copies of articles https://unpaywall.org/

Would be great to capture full text for academic publications where its available.

pirate:
Parsing DOIs out of PDF/HTML content: https://github.com/MicheleCotrufo/pdf2doi

Found good candidate extractor dependencies to try for all the major free scientific paper DBs:

- https://github.com/Tishacy/SciDownl
- https://github.com/viown/libgen-dl
- https://github.com/unpywall/unpywall
- https://openaccessbutton.org/?from=integration&id=
[...]

PR #1450, opened 2024-06-06
`papers-dl` support

benmuth:
# Summary

This PR adds basic support for the [papers-dl](https://pypi.org/project/papers-dl/) extractor for downloading scientific papers. 

At this point, `papers-dl` only supports Sci-Hub, but it is planned to support other repositories. 

Current known issues with `papers-dl` include performance and reliability. Searching known Sci-Hub aliases for a paper is slow and it occasionally causes timeouts during an ArchiveBox add job. There are also other occasional failures, even when doing a repeat search for a paper you've already found. Potential problems with captchas are also currently unaddressed.

`papers-dl` supports fetching multiple types of paper identifiers, but I still need to add parsing for those to ArchiveBox. At this moment only URLs work. I also still need to add an integration test.

# Related issues

#720 

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
- I'm on a road trip for the next couple weeks, but very excited about this PR! Will merge it before DWeb

benmuth:
`papers-dl` now searches all providers for papers concurrently, so performance shouldn't be a major concern anymore.

Angeeee158 (approved):

farooqkz:
oh I'm very much excited about this!

pirate:
I added papersdl support on dev! thanks again!

Issue #1453, opened 2024-06-10
Feature Request: Ability to restrict search to specific fields


sclu1034:
## Type 

 - [ ] General question or discussion
 - [ ] Propose a brand new feature
 - [x] Request modification of existing behavior or design

## What is the problem that your feature request solves
The existing search field only ever filters across all possible fields (content AND title AND url etc.). But sometimes I want to search for entries where I know that a word shows up in the title or URL, but the result list gets bloated by all the pages where this word shows up in the content instead.

## Describe the ideal specific solution you'd want, and whether it fits into any broader scope of changes
Implement a way to restrict the search to specific fields. For my use case, it would be fine to add a dropdown next to the search bar with options like `All`, `Title`, `Url`, `Content`, where `All` would be the default and current behavior of a searching everything, while the other options restrict the search to just that value.

## What hacks or alternative solutions have you tried to solve the problem?
None so far.
I suppose I could go into the archive storage, figure out how the metadata is saved and cobble together a script. But my instance is running inside Docker on Unraid, and both of these technologies aren't made for interactive shell usage.

## How badly do you want this new feature?

 - [ ] It's an urgent deal-breaker, I can't live without it
 - [x] It's important to add it in the near-mid term future
 - [ ] It would be nice to have eventually

---

 - [ ] I'm willing to contribute [dev time](https://github.com/ArchiveBox/ArchiveBox#archivebox-development) / [money](https://github.com/sponsors/pirate) to fix this issue
 - [x] I like ArchiveBox so far / would recommend it to a friend
 - [ ] I've had a lot of difficulty getting ArchiveBox set up

PR #1457, opened 2024-06-19
Implement searching individual fields

sclu1034:
# Summary

This adds a rudimentary implementation of #1453.
So far, this only changes the search box on the public view, as the custom form and corresponding logic there were relatively simple to track down and change.
As far as I can tell, the one on the admin page uses Django's built-in functionality instead, and I didn't see any immediate way to change that logic.

The current list of supported fields is mostly just based on the placeholder in the search input, but adding more metadata fields should be trivial.
Another interesting change could be to have the `Metadata` option run ripgrep, but restricted to `index.json` via glob (or the equivalent for other backends), but that would require a change in the search backends' `search()` function signature.

# Related issues

#1453

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture


pirate:
Works great, thanks!

image

PR #1627, opened 2024-12-25
Fix typo in timestamp scale factor

1over137:
I noticed that importing using timestamps doesn't seem to work. This is probably a typo. The timestamps are in microseconds (divide by 1e6). There is an extra zero there.

### DIFF
--- a/archivebox/parsers/generic_json.py
+++ b/archivebox/parsers/generic_json.py
@@ -26,7 +26,7 @@ def jsonObjectToLink(link: str, source: str):
     ts_str = str(datetime.now(timezone.utc).timestamp())
     if link.get('timestamp'):
         # chrome/ff histories use a very precise timestamp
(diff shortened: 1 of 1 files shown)


pirate:
Where are you importing from? MacOS/windows Firefox/Chrome/Safari are all different

1over137:
Actually I manually processed the files using a different scripts. The csv from python browser_history module uses a human readable format. 

But regardless: Firefox's places.sqlite is in microseconds relative to unix epoch, and Chrome's History (sqlite db) is in microseconds relative to 1601-01-01. Perhaps this should be handled by an export script, rather than the json parser, though. 

See:

https://stackoverflow.com/questions/20458406/what-is-the-format-of-chromes-timestamps

https://support.mozilla.org/en-US/questions/835204

pirate:
merged. thanks for contributing the fix!

Issue #1646, opened 2025-02-01
Bug: All timers/progress bars timeout when archiving from REST API


benmuth:
### Provide a screenshot and describe the bug

When using the extension (in the `redesign` branch, using the REST API) to archive any URL, all timers/progress bars take the maximum amount of time. Parsers take 240 seconds, most extractors take 60 seconds, the media extractor takes an hour (I think, I didn't wait around to find out). This doesn't happen when running `archivebox add` through the CLI, only the REST API.

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/0e7d032f-45d2-4aca-be62-45912e9ffc21" />

I kind of figured out what's going on, but not why. In `archivebox/logging_util.py`, `TimedProgress.end()` tries to `terminate` the `progress_bar` process. For some reason (busy writing to stdout?), the process ignores the terminate, then the `join()` call after the terminate just blocks until the `progress_bar` function finishes execution.

Adding an explicit signal handler to the beginning of the `progress_bar` function seems to fix the problem:
```
    def handle_term(signum, frame):
        sys.exit(0)

    signal.signal(signal.SIGTERM, handle_term)
```

That works, but I'm not sure if there's a better solution. Should I open a PR with this fix against dev?

### Steps to reproduce

```markdown
1. run docker run -it -p 8000:8000 \                                                                                    
                       -v $PWD/data:/data \
                       -v $PWD/archivebox:/app/archivebox \
                       archivebox server 0.0.0.0:8000 --debug
2. load the redesign folder on the redesign branch of the archivebox-browser-extension as an extension.
3. try to archive a webpage by clicking on the extension in the Chrome menubar.
```

### Logs or errors

```shell
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [2025-02-01 11:35:03] ArchiveBox v0.8.5rc53: archivebox server 0.0.0.0:8000                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
[+] Starting ArchiveBox webserver...
    > Starting ArchiveBox webserver on http://0.0.0.0:8000
    > Log in to ArchiveBox Admin UI on http://0.0.0.0:8000/admin
    > Writing ArchiveBox error log to ./logs/errors.log
Performing system checks...

System check identified no issues (0 silenced).
February 01, 2025 - 11:35:04
[...]

PR #1649, opened 2025-02-03
Kill the timer process if it doesn't properly terminate.

benmuth:
# Summary
This tries to fix the problem described in #1646. I ended up using `kill()` instead of the solution presented there because it turns out the behavior varied from run to run, and I couldn't get that solution to work consistently. Sending a kill signal was the simplest and most consistent fix. There may be some downsides I'm missing.

# Related issues
#1646 

# Changes these areas

- [x] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/archivebox/misc/logging_util.py
+++ b/archivebox/misc/logging_util.py
@@ -155,9 +155,10 @@ class TimedProgress:
                 time.sleep(0.1)
                 # sometimes the timer doesn't terminate properly, then blocks at the join until
                 # the full time has elapsed. sending a kill tries to avoid that.
-                if self.p.is_alive():
+                try:
                     self.p.kill() 
-                self.p.join()
+                except Exception:
+                    pass
 
 
                 # clear whole terminal line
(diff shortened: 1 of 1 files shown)


archivebox/misc/logging_util.py
```diff
@@ -152,6 +152,11 @@ def end(self):
                 except BaseException:                                           # lgtm [py/catch-base-exception]
                     pass
                 self.p.terminate()
+                time.sleep(0.1)
+                # sometimes the timer doesn't terminate properly, then blocks at the join until
+                # the full time has elapsed. sending a kill tries to avoid that.
+                if self.p.is_alive():
+                    self.p.kill() 
                 self.p.join()
```
pirate:
```suggestion
                time.sleep(0.1)
                # sometimes the timer doesn't terminate properly, then blocks at the join until
                # the full time has elapsed. sending a kill tries to avoid that.
                try:
                    self.p.kill() 
                except Exception:
                    pass
```
better to ask forgivness than permission (avoids TOCTTOU bug)

pirate:
thanks!

PR #1687, opened 2025-07-14
add optional ipfs support

F-Node-Karlsruhe:
# Summary

Add an optional IPFS storage backend support. The archived content can be additionally or only be stored in  IPFS and the database only stores the IPFS hast as reference

# Related issues

-

# Changes these areas

- [ ] Bugfixes
- [x] Feature behavior
- [x] Command line interface
- [x] Configuration options
- [x] Internal architecture
- [x] Snapshot data layout on disk

### DIFF
--- a/README.md
+++ b/README.md
@@ -897,6 +897,79 @@ Each snapshot subfolder <code>data/archive/TIMESTAMP/</code> includes a static <
 </details>
 <br/>
 
+## IPFS Storage Support
+
+ArchiveBox includes optional IPFS (InterPlanetary File System) storage support. When enabled, archived files are automatically uploaded to IPFS in addition to being saved locally, providing decentralized storage and permanent content-addressed URLs.
+
+<br/>
+<details>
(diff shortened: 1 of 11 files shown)


F-Node-Karlsruhe:
One could optinally spin up an IPFS node in docker compose

pirate:
Woah this is interesting, sorry I didn't see this sooner and comment!

Personally my preference is to keep storage-layer concerns outside of the core codebase. e.g. keep archivebox data in a normal FS but have a separate container periodically scan and upload all new data files using rclone or some other CLI tool.

The ArchiveBox codebase is already getting too big for me to maintain solo in my spare time, so I am unlikely to merge this at the moment. My apologies!

If you want to move some of this logic into a separate repo e.g. `ArchiveBox/archivebox-ipfs` that could be run as a standalone container in docker that shares access to `data/` that would be awesome. I'd be happy to document that repo as an official solution for IPFS.

PR #1710, opened 2025-11-13
Add deno and update yt-dlp

davidwilemski:
# Summary

yt-dlp now requires a js runtime and suggests deno >=2.0 https://github.com/yt-dlp/yt-dlp/issues/15012

In addition to updating the yt-dlp package there is now also a recommended external javascript blob that needs to be installed. See the [yt-dlp wiki] for install options. Here, I've added the `[default]` dependency group which will also install [yt-dlp-ejs](https://pypi.org/project/yt-dlp-ejs/) containing that blob.

[yt-dlp wiki]: https://github.com/yt-dlp/yt-dlp/wiki/EJS

This ensures yt-dlp continues to work. I was seeing some intermittent issues with the media extractor before this. I've tested this branch by building a docker image and running it against my archivebox install and confirming that some URLs now archive media properly.

I don't know how often the `yt-dlp-ejs` package will end up needing to be updated which may or may not be an argument in favor of eventually having an option to use the option to run the external js [from npm](https://github.com/yt-dlp/yt-dlp/wiki/EJS#option-2-enable-ejs-script-downloads-from-npm) but for now this sticks with baking an immutable docker image.

I also wasn't familiar with pdm but to update the pdm lock file I just ran `pdm lock` which seems to have worked but also updated quite a few other dependencies in that lockfile.

# Related issues

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@postlight/parser": "^2.2.3",
+        "deno": "^2.5.6",
         "readability-extractor": "github:ArchiveBox/readability-extractor",
         "single-file-cli": "^1.1.54"
       }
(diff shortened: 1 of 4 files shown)


pirate:
We should make sure to use a single JS env for both single-file-cli and yt-dlp, otherwise archivebox will install 2 redundant JS runtimes.

pirate:
I fixed this in dev, archivebox now handles npm stuff and passes NODE_BINARY to the yt-dlp plugin, which uses the `--js-runtimes node:{NODE_BINARY}` arg to enable it.

PR #1760, opened 2026-01-31
FIX: docker build

pellaeon:
# Summary

This PR fixes the docker image build. Also fixes the uuid7 not found error on the first run of `archivebox init`.

---
## Summary by cubic
Fixes the Docker image build and the uuid7 error on first init. We now use uv-managed Python 3.13 and patch uuid.uuid7 before Django migrations.

- **Bug Fixes**
  - Docker: switch to uv-managed Python, create venv with uv --python, skip version check at build, and start with --init.
  - UUID7: add uuid_compat, import it early, and monkey-patch uuid.uuid7 on <3.14 to keep migrations working.

- **Dependencies**
  - Bump Python to 3.13.
  - Require uuid_extensions on Python <3.14.

<sup>Written for commit 9aa4f0de587d9cbe5c1e20155295b0353dd4f5a9. Summary will update on new commits.</sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,7 +69,7 @@ ENV TZ=UTC \
     npm_config_loglevel=error
 
 # Language Version config
-ENV PYTHON_VERSION=3.12 \
+ENV PYTHON_VERSION=3.13 \
     NODE_VERSION=22
 
 # Non-root User config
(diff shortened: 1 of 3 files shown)


pirate (approved):
thanks!

PR #1798, opened 2026-05-03
Update batch update semantics

TheChosenOne1997:
# Summary

# Related issues

# Changes these areas

- [ ] Bugfixes
- [ ] Feature behavior
- [ ] Command line interface
- [ ] Configuration options
- [ ] Internal architecture

---
## Summary by cubic
Optimized bulk snapshot updates and reindexing with real batching, atomic transactions, and safer resume logic. Added a `--preserve-save-hooks` flag to control save behavior, reducing DB contention and making large updates more reliable.

- **New Features**
  - `--preserve-save-hooks` CLI flag to keep model `save()` hooks/signals when queueing.
  - Reindex runs search plugins per batch (`batch_size`) and reports `reindexed` counts.
  - Resume accepts UNIX timestamps and strings; numeric values filter by `bookmarked_at` cutoff.
  - Phase 1 migration prefers fast directory move via `os.rename()` with EXDEV fallback.

- **Refactors**
  - Phase 2 and filtered updates use atomic batch transactions; default to `QuerySet.update()` when not preserving hooks.
  - Reindex prefetches latest ArchiveResult per snapshot/plugin to reset retries without N×M queries.
  - Only reconcile when a snapshot directory exists; fix invalid `current_step`; preserve title changes; improved progress logs and numeric folder resume in archive draining.
  - Dependencies updated in `uv.lock`: `abx-dl` 1.10.30, `abx-plugins` 1.10.30, `abxpkg` 1.10.6, `uv` 0.11.8 (now pinned from PyPI).

<sup>Written for commit 28f05ad18665b65de5b520a5ac425297979e9a64. Summary will update on new commits.</sup>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ out/
 tmp/
 data/
 data*/
+users/
 archivebox/tests/data/
 archive/
 output/
--- a/archivebox/cli/archivebox_update.py
+++ b/archivebox/cli/archivebox_update.py
@@ -435,7 +435,7 @@ def drain_old_archive_dirs(resume_from: str | None = None, batch_size: int = 100
                 transaction.commit()
 
                 # Cleanup: delete old dir and create symlink
-                if old_dir.exists() and old_dir != new_dir:
+                if old_dir != new_dir:
                     snapshot._cleanup_old_migration_dir(old_dir, new_dir)
 
                 stats["migrated"] += 1
--- a/archivebox/cli/archivebox_update.py
+++ b/archivebox/cli/archivebox_update.py
@@ -2,9 +2,11 @@
 
 __package__ = "archivebox.cli"
 
+import errno
 import os
 import time
 
+from itertools import islice
 from typing import TYPE_CHECKING, Any
 from collections.abc import Callable, Iterable
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


TheChosenOne1997 (commented):
This fixup addresses a Phase 1 migration edge case where a successful os.rename() move causes the legacy archive/{timestamp} path to disappear, and the previous old_dir.exists() guard prevented _cleanup_old_migration_dir(old_dir, new_dir) from running—so the legacy symlink was never recreated


Change: The cleanup condition now runs whenever old_dir != new_dir (removing the old_dir.exists() requirement), ensuring the legacy symlink recreation path is executed after both rename-based migrations and copy-fallback migrations.

Test: Added a regression test (test_drain_old_archive_dirs_recreates_legacy_symlink_after_rename) that sets up a legacy snapshot directory and verifies drain_old_archive_dirs() migrates it and recreates the legacy path as a symlink pointing to the new 0.9.0 storage location.

PR update: Addressed the Phase 1 migration issue flagged by cubic: successful os.rename() could skip _cleanup_old_migration_dir() due to the old_dir.exists() guard, preventing recreation of the legacy archive/{timestamp} symlink. Fixed by running cleanup whenever old_dir != new_dir, and added a regression test (test_drain_old_archive_dirs_recreates_legacy_symlink_after_rename) to verify the legacy path becomes a symlink to the new 0.9.0 storage location. ✅ uv run pytest archivebox/tests/test_update.py -q → 8 passed.

Goal
Improve reliability and scalability of archivebox update / update --index-only for large archives by reducing contention and making update/index operations behave correctly under batching and resume flows

Key behaviors targeted (high level)

Bulk maintenance correctness: archivebox update is documented to perform Phase 1 legacy dir draining/migration and Phase 2 DB scan/re-queue behavior.

Legacy compatibility: ArchiveBox’s storage layout expects backwards-compatible symlinks under the legacy archive/ path after migration

Fixup included (what changed + why)
A review bot flagged a correctness issue in Phase 1 migration:

“Successful rename skips the migration cleanup path, so the legacy archive path never gets recreated as a symlink.

Root cause: When migration uses os.rename(), the original old_dir disappears; the previous cleanup guard required old_dir.exists(), so cleanup/symlink recreation could be skipped.Fix (in the fixup commit):

Changed cleanup condition from if old_dir.exists() and old_dir != new_dir: → if old_dir != new_dir: so _cleanup_old_migration_dir(old_dir, new_dir) always runs when paths differ, ensuring the legacy archive/{timestamp} symlink can be recreated even after a successful rename.
Added regression test test_drain_old_archive_dirs_recreates_legacy_symlink_after_rename verifying drain_old_archive_dirs() recreates the legacy symlink and it resolves to the new 0.9.0 storage location.

This aligns with the documented Phase 1 intent to “leave symlink in archive/ pointing to new location” after draining old directories.
Validation (as reported in PR workflow)

Fixup commit adds test coverage and updates migration cleanup behavior; the test is designed to catch the rename-path regression going forward.

pirate:
All this should be fixed on `:dev` now. I added many DB optimizations and guards to the migration process.

Please try the latest `dev` branch and open new issues if you find problems.

PR #1818, opened 2026-06-06
Render yt-dlp media inline in snapshot cards

FellowTraveler:
## Summary
- render archived yt-dlp video/audio outputs inline in snapshot cards using native browser media controls
- prefer playable yt-dlp media over thumbnails/subtitle/metadata files when selecting the primary output path
- include media type metadata in yt-dlp media helper output and keep links to all downloaded media files

## Tests
- pytest archivebox/tests/test_ui_admin_snapshot.py -q

---
## Summary by cubic
Render archived `ytdlp` videos and audio inline in snapshot cards using native media controls, only for browser‑playable formats. Prefer browser‑playable media for the main output and keep links to all downloaded files.

- **New Features**
  - Show an inline player for `ytdlp` results only when the file is browser‑playable (video first, then audio).
  - Prefer browser‑playable video/audio over thumbnails, metadata, and non‑playable containers for the main output.
  - Add media type flags (video/audio), sort with browser‑playable first then larger files, and expand media/subtitle extensions.
  - List links to all media files below the player.

<sup>Written for commit 2a0131b7abd7473b4ec1ae8feda487d2afcd9803. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ArchiveBox/ArchiveBox/pull/1818?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/archivebox/core/models.py
+++ b/archivebox/core/models.py
@@ -4291,12 +4291,36 @@ class ArchiveResult(ModelWithDeleteAfter, ModelWithOutputDir, ModelWithNotes):
                 if Path(candidate).name.lower() == preferred_name:
                     return candidate
 
-        ext_groups = (
-            (".html", ".htm", ".mhtml", ".mht", ".pdf"),
-            (".png", ".jpg", ".jpeg", ".gif", ".webp", ".svg", ".ico"),
-            (".json", ".jsonl", ".txt", ".md", ".csv", ".tsv"),
-            (".mp4", ".webm", ".mp3", ".opus", ".ogg", ".wav"),
-        )
+        plugin_lower = (plugin_name or "").lower()
+        if plugin_lower in ("ytdlp", "yt-dlp", "youtube-dl"):
+            ext_groups = (
+                (".mp4", ".webm", ".m4v", ".ogv"),
+                (".mp3", ".m4a", ".aac", ".opus", ".ogg", ".wav", ".flac"),
+                (
+                    ".mkv",
+                    ".mov",
+                    ".avi",
+                    ".flv",
+                    ".wmv",
+                    ".mpg",
+                    ".mpeg",
+                    ".ts",
+                    ".m2ts",
+                    ".mts",
+                    ".3gp",
+                    ".3g2",
+                ),
+                (".html", ".htm", ".mhtml", ".mht", ".pdf"),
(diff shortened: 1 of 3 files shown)


pirate:
I tried this and decided against it a few months ago actually, unfortunately it doesn't work great when any media files are large, it blows up the pageload because the browser immediately tries to fetch a 700mb+ file potentially.

If we were to add this it would need to be in the plugin template, not in archivebox core: https://github.com/ArchiveBox/abx-plugins/blob/main/abx_plugins/plugins/ytdlp/templates/card.html

And it would need some logic to prevent autoloading for massive files + make sure the embed frame is marked with the right attributes so it lazy-loads / defers until after the rest of the page is loaded. If you need python code to supply the view, that would belong in a views.py file in the plugin that gets imported into archivebox.

FellowTraveler:
> I tried this and decided against it a few months ago actually, unfortunately it doesn't work great when any media files are large, it blows up the pageload because the browser immediately tries to fetch a 700mb+ file potentially.
> 
> If we were to add this it would need to be in the plugin template, not in archivebox core: https://github.com/ArchiveBox/abx-plugins/blob/main/abx_plugins/plugins/ytdlp/templates/card.html
> 
> And it would need some logic to prevent autoloading for massive files + make sure the embed frame is marked with the right attributes so it lazy-loads / defers until after the rest of the page is loaded. If you need python code to supply the view, that would belong in a views.py file in the plugin that gets imported into archivebox.

FYI fixing this now.

FellowTraveler:
Thanks, agreed. I revised this PR so ArchiveBox core no longer renders the yt-dlp player directly. It now only exposes media metadata to plugin templates and fixes yt-dlp output selection to prefer browser-playable media before thumbnails / non-browser containers.\n\nI moved the actual player UI into a companion abx-plugins PR: ArchiveBox/abx-plugins#34. That template uses click-to-load buttons and only creates the video/audio element after user interaction, so large archived media files are not fetched during page load. Non-browser-playable media remains a regular file link.

PR #1820, opened 2026-06-08
FIX: error 500 when trying to pause a crawl

pellaeon:
Culprit:
Django admin localizes datetimes before rendering, and positive UTC offsets can overflow year 9999.

# Summary

Currently, if the system uses a timezone offset > UTC+0 , the max value used by the crawl job model can overflow when django is trying to render, causing internal server error.

This fix simply lowers the max value so that it does not overflow and adds a test.

---
## Summary by cubic
Fixes HTTP 500 when pausing a crawl by capping the retry timestamp to a safe max value so Django admin timezone localization doesn’t overflow on positive UTC offsets.

- **Bug Fixes**
  - Set `RETRY_AT_MAX` to `datetime(9999, 1, 1, tzinfo=UTC)` instead of `datetime.max`.
  - Added a test ensuring localization in `Pacific/Kiritimati` keeps the year at 9999.

<sup>Written for commit 766989eca32a25b88a01a82b1adf7df8c1ccbfa2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ArchiveBox/ArchiveBox/pull/1820?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/archivebox/tests/test_crawl_pause.py
+++ b/archivebox/tests/test_crawl_pause.py
@@ -5,6 +5,9 @@ import time
 
 import pytest
 import requests
+from django.utils import timezone
+
+from archivebox.workers.models import RETRY_AT_MAX
 
 from .conftest import (
     build_test_env,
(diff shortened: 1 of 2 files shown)


pirate (approved):

pirate:
I pulled the RETRY_AT_MAX fixes into `dev` in a separate commit, thanks!

PR #1871, opened 2026-09-04
Test immediate OpenCode events and real terminal ticket handshakes

pirate:
Add regressions for real upstream SSE first-event delivery and OpenCode protocol-header forwarding. Extend the native-browser terminal test to obtain and use the same connection ticket as the actual UI. The SSE regression failed against plugin 1.12.228; the terminal regression failed with a real 403 PtyForbiddenError against 1.12.229. The dependency cascade now pins published abx-plugins 1.12.230 and abx-dl 1.12.267. Validation without source overrides: uv run --locked --no-sources pytest archivebox/tests/test_opencode_agent.py archivebox/tests/test_opencode_browser.py -q --tb=short: 29 passed in 336.83s. Live DigestBox browser verification also showed a streamed model reply without reload, actual shell tools, provider authorization setup, and pwd returning /data in the embedded terminal. No mocks or relaxed assertions.


pirate:
Full real agent/browser suite passes: 29 tests in 354.49s using the merged plugin protocol-header fix. Published plugin 1.12.229 separately passes all three SSE tests without source overrides. Fixed the new Cubic cleanup comment by saving the DELETE response in finally and asserting its status afterward, preserving both the primary failure and the cleanup assertion. Awaiting the plugin release cascade before final no-source test and merge.

